Issue #90, opened 2026-05-27
roadmap: "85% custom-CSS reduction" plan — MobileFullscreen, SearchField, Variants/Density/Shape, Theme.Customize builder, Splash Kit


Brain2k-0005:
From production-consumer roadmap analysis. Consumer estimates ~85% of their custom CSS/JS could be eliminated with a focused set of Lumeo improvements. Filing as a roadmap-level tracking issue. Several items overlap with already-filed issues; cross-referenced below.

## Already shipped or in flight

| Item | Status |
|---|---|
| AutoFocus on form inputs | ✅ shipped in #79 (six inputs) + #84 (Combobox/Select/DatePicker). Consumers can drop their `JS.InvokeVoidAsync("eval", ...)` calls. |
| `<Image Class="h-14 w-auto" />` collapsed | ✅ fixed in #85 — consumers can use `<Image>` again instead of native `<img>` for sized logos. |
| `<SheetBody>` component | 🟡 filed in #86 high-priority section |
| EditForm-in-SheetContent flex hell | 🟡 filed in #86 |
| FileUpload Variant=Trigger | 🟡 filed in #86 |

#### `OverlayOptions.MobileFullscreen = true` (bool property)
Today consumers chain `max-md:!h-full max-md:!w-full ...` Tailwind overrides to make a Sheet / Dialog full-screen on mobile while keeping the side-sheet / dialog shape on desktop. A single boolean on the Overlay options that flips this internally would replace whole helper classes consumers build. One real consumer already maintains a `ResponsiveDialog` preset with this exact override chain.

#### `<Input Variant="Search">` or `<SearchField>`
Consumers add `min-w-[180px] max-w-md` and a manual search icon to every search field — 5+ repeats in one consumer project. A first-class search variant with the icon + sensible width defaults baked in.

#### `<UploadTrigger>` / `<FileButton>`
For file-pick-only flows (no list, no preview), consumers drop back to native `<label for>` because `<FileUpload>` always renders its own chrome. A standalone trigger component would close that loop.

#### Smaller-than-Xs size token
`text-[10px]` / `text-[11px]` arbitrary-value Tailwind escape hatches today on Avatar / Badge dense modes. Add a `Lumeo.Size.Xxs` token (or per-component `Size="Xxs"`).

#### Variants instead of `!override`-hell
[...]

Issue #89, opened 2026-05-27
bug-triage: DatePicker double-calendar-icon on mobile (needs repro) + OtpInput multi-paste already works


Brain2k-0005:
From production-consumer feedback. Two reported bugs that need triage before fixing.

## 1. DatePicker mobile: double calendar icon (Prefix + Trigger)
**Symptom (reported)**: when consumer composes DatePicker, two `Lucide.Calendar` icons appear next to each other on mobile (<500px), making the input visually noisy.

**Triage**: DatePicker.razor renders one `<Blazicon Svg="Lucide.Calendar" />` at the start of each trigger branch (line 45 single, line 153 masked-input, line 194 range). There's no `Prefix` parameter on DatePicker — so the "double icon" must come from a consumer-side composition (e.g. wrapping in a custom Input-like component that adds its own icon prefix, or stacking with `<Input Prefix=...>`).

**Need**: a concrete repro Razor snippet from the consumer. If it turns out to be the consumer doubling up icons inadvertently, this becomes a docs / `IconPosition` parameter on DatePicker. If the icon shows up twice in pure single-component usage, that's a real bug.

## 2. OtpInput — already addressed, closing note
The feedback claimed "OtpInput unterstützt keinen Multi-Cell-Paste oder gezieltes FocusAsync()". On verification:
- **Multi-cell paste**: already works. `OtpInput.OnAfterRenderAsync` calls `Interop.RegisterOtpPaste(_baseId, Length, HandlePaste)` (line 145 / line 155). Pasting "123456" into any digit cell distributes the digits across cells. Test in `tests/Lumeo.Tests/Components/OtpInput/OtpInputTests.cs`.
- **`FocusAsync()`**: shipped in #79 (along with `[Parameter] bool AutoFocus`).

So #2 is already done — just flagging here so the feedback list closes cleanly.

Brain2k-0005:
Root cause fix shipped in #110 / 3.4.0: the `AllowKeyboardInput` variant rendered both a leading `Lucide.Calendar` icon AND a trailing `Lucide.CalendarDays` button — most visible on narrow mobile widths. Removed the leading icon in that branch; trailing button covers visual + click affordances. Other variants unchanged. Closing as completed.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #88, opened 2026-05-27
docs: gaps — Long forms in Sheets, service vs markup overlays, mobile Sheet patterns, Blazor-WASM lifecycle, theme overrides


Brain2k-0005:
From production-consumer feedback. Five docs pages a real consumer wished existed.

## 1. "Long forms inside a Sheet"
Concrete recipe for: scrollable body + sticky footer + focus-rings-not-clipped on inputs near the edges. Reported manually adding `overflow-y-auto px-1 -mx-1` in 13 places; the docs don't show the pattern. Pair with the `<SheetBody>` proposal in #86 — even before that ships, a docs page would unblock consumers.

## 2. "Service-driven vs markup-driven overlays" — when to use which
`<Sheet @bind-Open>` vs `OverlayService.ShowSheetAsync` — the docs touch on both but don't draw the boundary clearly. Markup-driven for shells / persistent UI, service-driven for ad-hoc / fire-and-forget — but the trade-offs around EditForm composition, re-render lifetime, focus restoration, and which can return a value differ in non-obvious ways.

## 3. "Mobile-first Sheet patterns" — concrete Tailwind recipes
Sheets that are full-screen on mobile and side-sheet on desktop. Should show `max-md:!h-full` / `md:w-[400px]` / etc. + how `SheetSize` interacts with viewport-class overrides. Real consumers rebuild this pattern from scratch.

## 4. "Blazor WASM lifecycle limitations" doc page
A single canonical page covering:
- Why HTML `autofocus` doesn't fire (parse-time vs WASM-boot ordering) — now solved by `AutoFocus` param shipped in #79 / #84
- `OnInitialized` vs `OnAfterRender(firstRender)` — what runs where, what's available when
- Why JS interop must be in `OnAfterRender`, not `OnInitialized`
- `JSDisconnectedException` and the wrap pattern Lumeo uses internally
- Prerender vs hydration if relevant (the docs site uses Puppeteer prerender)

## 5. "Theme overrides" — single-token branding pattern
Consumers don't want a full theme — they want to override one or two CSS variables (brand color, radius) and keep the rest. Show the `theme.css` shape + how it composes with the default theme, plus which variables are stable vs which might shift between minor versions.

---

Each of these is a 1-page doc, examples copy-pasteable, ideally with the "patterns" filter on the existing patterns page so they're discoverable next to the examples that already exist.

Brain2k-0005:
[...]

Issue #87, opened 2026-05-27
dx: OverlayOptions typing, SheetSize/Lumeo.Size overlap, @bind-Value consistency, OverlayForm helper, MCP skill defaults


Brain2k-0005:
From production-consumer feedback. These are architectural papercuts rather than outright bugs — each one hit consumer code multiple times.

## 1. `OverlayOptions` is a fat class — split into type-specific records
Today there's a single `OverlayOptions` with both Sheet-specific (`SheetSize`, `Side`, ...) and Dialog-specific properties. Setting `SheetSize` on a Dialog overlay is a silent no-op. Better:

```csharp
public abstract record OverlayOptions { ... shared ... }
public sealed record SheetOptions   : OverlayOptions { Size, Side, ... }
public sealed record DialogOptions  : OverlayOptions { MaxWidth, ... }
```

Each `OverlayService.ShowXxxAsync` overload takes the type that matches.

## 2. `Lumeo.Size.Full` vs `Sheet.SheetSize.Full` enum overlap
3.x already unified Size/Side/Align/Orientation — big win. But Sheets still have their own `SheetSize` enum that overlaps semantically with `Lumeo.Size`. Either deprecate `SheetSize` in favour of `Lumeo.Size`, or document why both exist. Consumers reasonably try `Size="Lumeo.Size.Full"` first and have to learn the special case.

## 3. `@bind-Value` inconsistency across inputs
Some inputs accept the standard `@bind-Value` pattern, others need explicit `Value="..." + OnInput="..."` because they fire events at different points. NumberInput, TagInput, Combobox each behaved slightly differently in real consumer use. Standardise on `@bind-Value` everywhere — and where the value type needs a converter (NumberInput double?), provide the convention so the analyzer guides consumers.

## 4. Service-driven Overlays + `EditForm` is awkward
`OverlayProvider` renders its own shell. When the consumer-component passed to `ShowSheetAsync` contains an `EditForm` with a submit button, the flex layout breaks (submit button doesn't pin to bottom, scroll doesn't compose correctly). Either:
- Doc + example: "EditForm in service-driven overlay"
- Or ship an `<OverlayForm>` wrapper that handles the flex-col + min-h-0 + footer-sticky shape and is the recommended pattern.

## 5. MCP component skill hides non-obvious defaults
[...]

PR #110, opened 2026-05-27
release: 3.4.0 — OverlayForm, Splash Kit, Theme.Customize, DataGrid + DatePicker + docs (closes #67, #88, #89; #87/#90 partial)

### LINKED ISSUE #67: bug: DataGrid — aria-rowcount, PageUp clamp, server-mode selection, grouped RowIndex
From the deep static bug-scan of `src/Lumeo.DataGrid/**`. Grid is generally well-built — these are the residual issues worth addressing in one focused pass.

## HIGH — aria-rowcount off-by-one
`src/Lumeo.DataGrid/UI/DataGrid/DataGrid.razor:296`

`aria-rowcount="@(_displayedItems.Count + 1)"` assumes the header occupies a logical row. The ARIA grid pattern wants aria-rowcount to be **total rows including headers and data**, but the displayed-rows count + 1 (for "the header") double-counts because aria-rowcount semantics treat all `role="row"` elements (header + data) uniformly. Verify against spec and either drop the `+1` or include all header rows.

## HIGH — PageUp keyboard clamp wrong on row 0
`DataGrid.razor:743`

When the user presses PageUp at row 0, `Math.Max(0, row - KeyboardPageStep)` clamps to 0 — a no-op. Either wrap to header (-1) so focus lands on the columnheader band, or document the no-op. Trivial:

```csharp
var newRow = row - KeyboardPageStep;
if (newRow < 0) newRow = (row > 0 ? -1 : 0); // wrap to header, else stay
```

## MED — Selection desyncs in server-mode after page reload
`DataGrid.razor:1364-1373` — `ProcessClientData` cleans stale selections via reference equality. In server-mode where Items are fresh instances per fetch, selections by reference don't match across pages. Fix: introduce an optional `[Parameter] Func<TItem, object>? SelectionKeySelector` for ID-based equality, and use it across `_selectedItems` lookups.

## MED — In-flight server request stale-data race
`Lumeo.DataGrid/Services/DataGridServerService.cs:25-67`

The generation guard protects `IsLoading` but not the returned data. If the consumer's `OnServerRequest` callback resolves out-of-order, the older response can overwrite the newer one. Fix: response-sequencing guard, or document that consumers must drop responses with a stale request token.

## MED — RowIndex off-by-one in grouped mode for keyboard navigation
`Lumeo.DataGrid/UI/DataGrid/DataGridBody.razor:114-117`

In grouped rendering the loop counter `RowIndex` resets per group. `DataGridCell` uses `RowIndex` for keyboard nav — Home/End/PageUp/PageDown anchor to row 0 of the current group, not the grid. Currently mitigated because `Navigable` is disabled for grouped rows (line 128), but worth ratifying with a test so the mitigation doesn't quietly disappear.

## LOW — Several smaller items
- `DataGridHeaderCell.razor:311-314` — column-resize listener may leak on rapid `Resizable` flag toggle. Track registration state and `await` the unregister before re-register.
- `DataGridHeaderCell.razor:318-325` — `PositionFixed` for filter popover recomputed every render — should fire only on `IsFilterOpen` false→true transition.
- `DataGridCell.razor:130` — `CssClass` memoization key missing `IsBatchDirty`; cached styling persists across dirty toggles.
- `DataGrid.razor:3046` — `DisposeAsync` could be more defensive around `Interop.UnlockScroll()` (current catch is enough but inconsistent with other components' pattern).

## Severity / effort
2 HIGH (both trivial) + 3 MED · small each. The two HIGH a11y/keyboard fixes are sub-10-line each.

Brain2k-0005:
## Summary

Reframes the previously-#67/#89-only branch as **3.4.0** — adds OverlayForm, Splash Kit (Hero / FeatureGrid / FeatureItem / CTASection), Theme.Customize fluent builder, and five long-form doc guides. Closes the bulk of the remaining 5 open issues going into this release.

### #67 — DataGrid
- **PageUp clamp at row 0** — wraps to `-1` (header band) instead of clamping at `0`. `MoveFocusAsync` already accepts `row=-1`.
- **`SelectionKeySelector` parameter** — new optional `Func<TItem, object?>` so server-mode `Items` re-fetches don't wipe selections via reference miss. `IsRowSelected` falls back to a key scan when set.
- **`DataGridServerRequest` XML doc** — spells out the `CancellationToken` cancellation contract.
- `aria-rowcount` off-by-one + grouped `RowIndex` test verified FP (header is 1 row → `+1` correct; grouped rows are non-`Navigable`).

### #89 — DatePicker double calendar icon
Root cause: `AllowKeyboardInput` variant rendered BOTH a leading `Lucide.Calendar` AND a trailing `Lucide.CalendarDays` button. Removed the leading icon in that branch — the trailing button covers visual + click affordances. Other variants unchanged.

### #87 — DX papercuts
- **#87.4 `OverlayForm`** — opinionated `EditForm` wrapper baking in `flex-col + min-h-0 + sticky-footer` so long forms in service-driven Sheets/Drawers/Dialogs don't lose their layout. Header / Body / Footer slots; falls back to `DataAnnotationsFormValidator`.
- **#87.2 `SheetSize` vs `Lumeo.Size`** — addressed via the existing XML doc on `Size.cs` that already documents `Full` as a deliberate layout-intent exception (kept on per-component enums by design).
- **#87.1 `OverlayOptions` split** — deferred. Breaking change requires owner direction on migration story.
- **#87.3 `@bind-Value` consistency** — deferred. Audit needs a per-input behaviour pass; not in scope for one PR.
- **#87.5 MCP skill defaults** — config/data-only follow-up; needs a separate PR against `tools/lumeo-mcp/`.

### #88 — Docs gaps  (closed)
Five new guides under `docs/Lumeo.Docs/Pages/Docs/`, registered in `nav-config.json` between Accessibility and Changelog:
- `LongFormsInSheets`
- `ServiceVsMarkupOverlays`
- `MobileSheetPatterns`
- `BlazorWasmLifecycle`
- `ThemeOverrides`

⚠ Snippet API names (`SheetFooter`, `SnapPoints`, `ShowDragHandle`) were written illustratively and not exhaustively verified against current component signatures; review before publishing.

### #90 — Roadmap
- **#90.2 `Theme.Customize` builder** — fluent `ThemeBuilder` with `WithPrimary` / `WithBorderRadius` / `WithFontFamily` / `WithVariable`, emits a managed `<style id="lumeo-custom-theme">` block via `window.lumeoThemeCustom.apply` (replaces on each Apply → no stale-rule leak).
- **#90.3 Splash Kit** — `Hero`, `FeatureGrid`, `FeatureItem`, `CTASection` landing-page primitives.
- **#90.1 Variants/Density/Shape** — deferred. Touches every sized component (Input, Button, Select, Combobox, Avatar, Chip, …) → multi-day refactor pass; not in scope for one PR.
- (MobileFullscreen + SearchField already shipped in 3.3.0.)

## Bump
`Directory.Build.props` Version 3.3.0 → **3.4.0** (new public API surface: `OverlayForm`, `ThemeBuilder`, `Hero`, `FeatureGrid`, `FeatureItem`, `CTASection`, `DataGrid.SelectionKeySelector`).

## Test plan
- [ ] CI green
- [ ] DataGrid: server-mode `SelectionKeySelector` survives a page change
- [ ] DataGrid: PageUp on row 0 → focus moves to header row
- [ ] DatePicker: `AllowKeyboardInput=true` on mobile → exactly one calendar icon
- [ ] OverlayForm: long form inside a Sheet → body scrolls, footer pinned
- [ ] Splash Kit pages: Hero / FeatureGrid / CTASection render at all breakpoints
- [ ] Theme.Customize: `.WithPrimary("#3b82f6").ApplyAsync()` then `.ClearAsync()` round-trips cleanly

## Summary by CodeRabbit

* **New Features**
  * Added new landing page components: Hero, FeatureGrid, FeatureItem, CTASection, and OverlayForm.
  * DataGrid now supports key-based row selection for server-side data scenarios.
  * Theme customization API enables dynamic CSS variable overrides.

* **Bug Fixes**
  * Removed duplicate calendar icon in DatePicker when using keyboard input.

* **Documentation**
  * Added 5 new guides: Blazor WASM Lifecycle, Long Forms in Sheets, Service vs Markup Overlays, Mobile Sheet Patterns, and Theme Overrides.

* **Chores**
  * Version bumped to 3.4.0.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Brain2k-0005/Lumeo/pull/110?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Brain2k-0005:
@codex review

Addressed both P2 findings in `bb746137`:
- ToggleSelection now consults `FindSelectionMatch` (ref + key) before mutating, so a fresh server-mode wrapper deselects instead of duplicating.
- Stale-selection cleanup now keeps null-key items if they're still present by reference, preserving the documented "null key → reference equality" fallback in mixed scenarios.

---

Brain2k-0005:
@codex review

Addressed third P2 in `70e40684` — `WithFontFamily` now writes the raw `font-family` declaration (matching `ThemeService.SetFont`) instead of an unconsumed `--font-family` custom property.

---

Brain2k-0005:
@codex review

Addressed all 3 P2s from the previous round + 2 CI build errors in `2919d358`:

**CI build fixes**
- `OverlayForm.razor`: switched fallback to the framework's `<DataAnnotationsValidator />` — the previous `DataAnnotationsFormValidator` is a service class, not a component.
- `ServiceVsMarkupOverlays.razor:13`: escaped `@bind-Open` → `@@bind-Open` inside the `<code>` element so Razor stops parsing it as a directive.

**Codex P2 fixes**
- `DataGrid.FindSelectionMatch`: refactored to `(bool found, TItem stored)` — eliminates the `TItem?` sentinel that was non-null for value-type rows and made every unselected `int`/struct row report selected.
- `ThemeBuilder.WithVariable`: dropped the misleading "shorthand" implication in the XML doc — passing `"primary"` correctly emits `--primary` (which the bundled CSS does not consume); consumers must pass the full token name or use the typed `WithPrimary` / `WithRing` setters.
- `ThemeOverrides.razor`: bulk-renamed `--primary` / `--ring` / `--accent` / `--background` / `--foreground` / `--muted` / `--secondary` / `--border` / `--destructive` (and their `-foreground` pairs) to `--color-*` in both the docs table and the scoped demo so the live preview actually picks up the override.

---

Brain2k-0005:
@codex review

Addressed all 9 P2s from the previous round in `8ff8f94b`:

**Code**
- DataGrid header `AllSelected` / `SomeSelected` now probe each `DisplayedItem` via `Context.IsRowSelected` instead of comparing `_selectedItems.Count >= DisplayedItems.Count` — fixes the false all-selected state under `SelectionKeySelector` + server-mode + small `PageSize`.
- `ThemeBuilder.WithVariable` now throws `ArgumentException` when the name doesn't start with `--`. No more silently-unused `--primary` from the documented `WithVariable("primary", …)` shorthand path; failure is loud.

**Docs**
- `ServiceVsMarkupOverlays`: rewrote the service snippet against the real `IOverlayService.ShowSheetAsync(title, side, size, parameters, options)` signature; `result.Cancelled` + `result.GetData<T>()` instead of nonexistent `IsConfirmed`.
- `LongFormsInSheets`: `Side` / `Size` moved onto `SheetContent` (where they actually live); `Sheet` only owns the open/dismiss params.
- `MobileSheetPatterns`: deleted the Snap Points and `ShowDragHandle` sections — neither API exists on Sheet/SheetContent. Reframed swipe-to-close as opt-in (matches the `SwipeToClose=false` default). `SafeArea Bottom="true"` instead of nonexistent `Edges`. Responsive switching now uses `IResponsiveService.IsMobile` + `ViewportChanged` (no `Interop.MatchesMediaAsync` — that method doesn't exist). Added a new section showing the real `MobileFullscreen` overlay option for service overlays.

---

Brain2k-0005:
@codex review

Addressed all 8 P2s from `8ff8f94b` in `99f8f13c`:

**Code**
- DataGrid `SelectAll` / `ClearSelection`: now MERGE the current page under `SelectionKeySelector` instead of replacing. Header-checkbox actions on page 2 no longer wipe page-1 selections. Symmetric clear removes only visible rows when keyed.

**Docs**
- LongFormsInSheets: `Button.IsLoading` (not `Loading`); `AlertDialogOptions.IsDestructive` (not `DestructiveConfirm`); `result.Cancelled` (not `IsConfirmed`).
- ThemeOverrides: `--card` → `--color-card`, `--input` → `--color-input` in the variable table.
- ServiceVsMarkupOverlays: markup snippet wraps content in `<SheetContent>` and uses `Lumeo.Side` / `SheetContent.SheetSize`.
- BlazorWasmLifecycle: `RegisterClickOutside` real signature (string elementId, string? triggerId, Func<Task>) returning `ValueTask`; dropped the bogus `@ref` + `IDisposable` pattern.

**Registry**
- `Hero` / `FeatureGrid` / `FeatureItem` / `CTASection` / `OverlayForm` will be regenerated automatically by `.github/workflows/sync-mcp-registry.yml` on master push — that workflow watches `src/Lumeo*/UI/**/*.razor` and runs `Lumeo.RegistryGen`, then commits the regenerated `registry.json` and `components-api.json`. No manual regen needed before merge.

**Build**
- The one test failure on `8ff8f94b` was the flaky `ToastProvider_MaxVisible_2_With_3_Toasts_Shows_Only_2` (pre-existing, unrelated to this PR). Re-run should be green.

---

Brain2k-0005:
Status update on `dd907640`:

**Codex's 3 P2 repeats are stale re-anchors** — the findings describe code that's already fixed:

- `DataGrid.razor:640` finding text says *"DataGridHeader still derives the header checkbox state from `SelectedItems.Count >= DisplayedItems.Count`"* — but `DataGridHeader.razor:98-129` (since `8ff8f94b`) uses a per-row probe via `Context.IsRowSelected`. The Codex anchor is on the `SelectionKeySelector` parameter declaration, not on the actual logic line; the description references the pre-fix state.
- `DataGridHeader.razor:110` finding text says *"its change path still calls `SelectAll()` / `ClearSelection()`, where `SelectAll()` replaces `_selectedItems` with `_displayedItems`"* — but `DataGrid.razor:2292-2342` (since `99f8f13c`) now MERGES under `SelectionKeySelector` instead of replacing. Same stale-anchor pattern.
- `Hero.razor:7` — the registry regen is handled by `.github/workflows/sync-mcp-registry.yml` automatically on master push (it watches `src/Lumeo*/UI/**/*.razor`). Pre-merge regen would require running `Lumeo.RegistryGen` locally; the workflow run after merge produces a `chore(mcp-sync)` commit that ships the entries.

**CodeRabbit findings (11) — handled in `dd907640`:**

Applied:
- `ThemeBuilder` constructor + `Customize` factory now throw `ArgumentNullException` for null `IJSRuntime`.
- `FeatureItem.razor` moved from `src/Lumeo/UI/FeatureGrid/` to its own `src/Lumeo/UI/FeatureItem/` folder (CLAUDE.md folder convention).
- `OverlayForm.Validator` XML doc `cref` corrected: `DataAnnotationsFormValidator` → `DataAnnotationsValidator`.
- `lumeoThemeCustom.apply` / `.clear` now dispatch `lumeo:theme-changed` so chart modules and other consumers re-read CSS variables after a `ThemeBuilder` apply (matches the preset path).

Skipped with reason:
- "Apply `@namespace Lumeo` to docs pages" (5 findings) — false alarm. Those pages live under `docs/Lumeo.Docs/` which has its own `Lumeo.Docs.Pages.*` namespace policy. The `@namespace Lumeo` + `Class` / `AdditionalAttributes` contract from CLAUDE.md applies to library components under `src/Lumeo/UI/`, not Razor pages in the docs site.
- The two analysis-only HashSet-comparer / ComponentInteropService probes from CodeRabbit had no concrete suggestion attached and the verified code is already correct.

---

Brain2k-0005:
@codex review

Latest commit `dd907640`. The three repeated findings from prior rounds (DataGrid select-all state, header SelectAll merge, Hero registry) — please re-verify against the current HEAD:

- `DataGridHeader.razor:103-128` — `AllSelected` / `SomeSelected` are per-row `Context.IsRowSelected` probes; no count comparison.
- `DataGrid.razor:2292-2344` — `SelectAll()` and `ClearSelection()` MERGE the current page into `_selectedItems` when `SelectionKeySelector` is set (no replace-with-displayed).
- Hero / Splash Kit registry: `sync-mcp-registry.yml` regenerates `registry.json` + `components-api.json` automatically on master push (the workflow watches `src/Lumeo*/UI/**/*.razor`). Pre-merge regen would require running `Lumeo.RegistryGen` from a dotnet-equipped runner; happy to do that as a separate commit if you'd prefer the entries to ship inside this PR rather than via the post-merge sync.

---

PR #139, opened 2026-05-30
perf(docs): defer hero BarChart's ECharts init off the critical path (IdleMount)

Brain2k-0005:
## Summary

Middle-path landing-perf change (per discussion): keep the live-demo hero, but take the **biggest page-controllable TBT contributor — ECharts — off the first-paint critical path.**

### The lever
The hero "Revenue by package" `BarChart` is above the fold, so `LazyRender` can't help (visible → mounts immediately, like Codex flagged for the DataGrid). Its ECharts CDN fetch (~1 MB) + init + first render run during the post-WASM-boot re-render burst, which is the largest slice of landing TBT we can actually move.

### What changed
- **New `IdleMount` helper** (`docs/Shared/IdleMount.razor`): renders a cheap `Placeholder` immediately, then swaps in `ChildContent` on `requestIdleCallback` (`window.lumeo.onIdle` added to `docs.js`). Mounts immediately during prerender (snapshot stays complete) and falls back to immediate mount when JS is unavailable.
- **BarChart wrapped in `IdleMount`** with a pure-CSS stacked-bar placeholder built from the *same* `_barSeries` data, so it matches the real chart's shape and the swap is layout-stable. ECharts now initialises during idle time.

### What I deliberately did NOT do (and why)
The plan also floated converting the 5 hero `BlurFade`s to CSS-only and gating `NumberTicker`s on visibility. On inspection:
- Hero `BlurFade`s are **prerender-friendly — already fully visible above the fold with no entrance animation on first load** (`lumeo.css:820-836`). Replacing them yields ~zero visual change and only marginal TBT (5 of 20 IntersectionObserver registrations; `Lumeo.Motion` still loads for NumberTicker/ShimmerButton/Marquee). Low ROI, fiddly markup surgery — held.
- `NumberTicker`s are above-the-fold and *are* the KPI "wow" — deferring their animation doesn't help (immediately visible) and removing it loses the effect. Held.

So this PR is the one high-value, low-risk piece.

### Honest caveat on measurement
The container's headless Lighthouse TBT is too noisy to prove the delta (±2.5s run-to-run, larger than the effect). The change is logically sound — ECharts init moves out of the critical window — but the win needs **real-browser DevTools / CI Lighthouse** to quantify. FCP/LCP/visual are unchanged by construction.

## Test plan
- [x] `dotnet build docs/Lumeo.Docs -c Release` succeeds
- [ ] CI build-and-test + e2e
- [ ] Real-browser verification on the preview deploy (TBT delta; confirm chart swaps in correctly and placeholder matches)

---
_Generated by [Claude Code](https://claude.ai/code/session_01LJmBFYeCdQj2G2epoZaRzQ)_

## Summary by CodeRabbit

* **New Features**
  * The "Revenue by package" chart now displays a loading skeleton placeholder while content loads, providing visual feedback during the loading phase.

* **Performance**
  * Chart rendering deferred until browser idle for faster initial page load times.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Brain2k-0005/Lumeo/pull/139?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/Lumeo.Docs/Pages/Home.razor
+++ b/docs/Lumeo.Docs/Pages/Home.razor
@@ -116,7 +116,33 @@
                         </Badge>
                     </Flex>
                     <div class="h-56 sm:h-64 lg:h-80 px-2 pb-3">
-                        <BarChart Categories="_barCategories" Series="_barSeries" Stacked="true" Height="100%" ShowLoadingSkeleton="false" />
+                        @* IdleMount keeps ECharts' CDN fetch + init off the first-paint critical
+                           path: a cheap CSS stacked-bar placeholder (built from the same data, so
+                           it matches the real chart's shape) shows immediately, then the live
+                           BarChart swaps in once the browser is idle. *@
+                        <IdleMount>
+                            <Placeholder>
+                                <div class="flex h-full items-end gap-1.5 sm:gap-2" aria-hidden="true">
+                                    @for (var i = 0; i < _barCategories.Count; i++)
+                                    {
+                                        var total = _barSeries.Sum(s => s.Values[i]);
+                                        var barPct = _chartMaxTotal > 0 ? total / _chartMaxTotal * 100 : 0;
+                                        <div class="flex-1 flex flex-col-reverse rounded-t-sm overflow-hidden"
+                                             style="height:@barPct.ToString("0.#", System.Globalization.CultureInfo.InvariantCulture)%">
+                                            @for (var s = 0; s < _barSeries.Count; s++)
+                                            {
+                                                var segPct = total > 0 ? _barSeries[s].Values[i] / total * 100 : 0;
+                                                var op = (0.85 - s * 0.28).ToString("0.##", System.Globalization.CultureInfo.InvariantCulture);
+                                                <div style="height:@segPct.ToString("0.#", System.Globalization.CultureInfo.InvariantCulture)%;background:hsl(var(--primary) / @op)"></div>
+                                            }
+                                        </div>
+                                    }
+                                </div>
+                            </Placeholder>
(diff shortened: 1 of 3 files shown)


Brain2k-0005:
Triage of the CodeRabbit review against the existing precedent in `docs/Lumeo.Docs/Shared/` — the sibling helper `LazyRender.razor`, which `IdleMount` mirrors 1:1:

| Finding | LazyRender precedent | Verdict |
| --- | --- | --- |
| Add `@namespace Lumeo` as line 1 | LazyRender has no `@namespace` directive | Skip |
| Inject `ComponentInteropService` instead of `IJSRuntime` | LazyRender does `@inject IJSRuntime JS` and calls `JS.InvokeAsync(...)` directly | Skip |
| Add `Class` + `AdditionalAttributes` with a single root element | LazyRender has only `Class` for its skeleton div, no `AdditionalAttributes`, no enforced single root | Skip |
| Defensive bounds on `_barSeries[s].Values[i]` (×2) | Data is hardcoded one screen apart from the placeholder and CodeRabbit labelled both as "💤 Low value" | Skip |

The coding guidelines CodeRabbit cited (`@namespace Lumeo` on every `.razor`, `ComponentInteropService` for JS interop, `Class`/`AdditionalAttributes` on every component) are CLAUDE.md rules **scoped to library components under `src/Lumeo/UI/`** — they exist to keep the public NuGet surface uniform. `IdleMount` is an internal docs-app helper in `docs/Lumeo.Docs/Shared/`, not a library component, and the established pattern next to it doesn't follow those rules either. Applying them here would diverge from the codebase, not align with it.

No changes pushed.

---

Issue #337, opened 2026-06-17
[audit] Component audit index — Lumeo vs Radix + shadcn (163 components, v3.13.2)


Brain2k-0005:
# Component audit — Lumeo vs Radix + shadcn (2026-06-17, v3.13.2)

Self-audit of **all 163 components** (142 core + 21 satellite) against the Radix primitive / shadcn equivalent (or closest analog: AntD, MudBlazor, assistant-ui, Magic UI, TipTap, ECharts, …). Each component has its own `component-audit` issue with verified bugs, a11y gaps, and missing features. Bugs already fixed in 3.13.0–3.13.2 were verified in source and excluded.

**Severity:** 1× P0 · 30× P1 · 60× P2 · 72× P3 · 163 total

## Cross-cutting (the real "becomes Radix" roadmap)

- **Nested-overlay portaling** — #172 (sub-content/popover-in-popover break in a transformed ancestor).
- **Icon sizing under Tailwind v4** — #173 (Blazicons unlayered `svg` rule beats utilities).
- **`asChild` polymorphism** — absent library-wide (more wrapper DOM, less composition than Radix).
- **`prefers-reduced-motion`** — none of the 10 Lumeo.Motion components honor it.
- **Keyboard/focus depth** — recurring gap in overlays/menus/collections vs Radix's WAI-ARIA defaults.

## P0 (1)

- #212 **PivotGrid** — row key join collides, corrupts aggregates

## P1 (30)

- #206 **Calendar** — dead ShowYearPicker, no keyboard/ARIA grid
- #198 **Combobox** — composition Disabled, portal, focus gaps
- #311 **ConsentBanner** — never dismissable if all-Required; prefs dialog not a real modal
- #224 **ContextMenu** — root menu has no viewport clamp
- #218 **Drawer** — no snap points, distance-only dismiss
- #223 **DropdownButton** — disabled bypass; not a split button
- #314 **FileManager** — stale children after Root swap; rows have no keyboard nav
- #287 **Image** — modal preview lacks focus trap; no srcset/placeholder
- #213 **Kanban** — presentational DnD, no payload or keyboard
- #296 **Link** — no disabled, no aria-current, no NavLink active
- #227 **NavigationMenu** — no keyboard nav, no trigger ARIA
- #230 **PopConfirm** — span trigger not keyboard-operable
- #219 **Popover** — no focus management, Esc unreachable
- #211 **QueryBuilder** — in/notIn single-value editor mismatch
[...]

Issue #320, opened 2026-06-17
[audit] RichTextEditor — solid TipTap wrapper; gaps vs TipTap kit


Brain2k-0005:
**Component:** `RichTextEditor` (`Lumeo.Editor`) · **vs** TipTap / Lexical
_Audit 2026-06-17 · Lumeo 3.13.2 source · benchmarked vs TipTap/Lexical._

### 🔴 Bugs / correctness
- None found. TipTap is bundled (`tiptap-bundle.js` ~1.4MB, not CDN). JS disposal is exemplary: `destroy(id)` removes named drop/paste handlers, destroys bubble menu + drag handle + editor (`rich-text-editor.js:1212-1226`); suggestion popups track scroll/resize listeners and `removeEventListener` on teardown (`:282-289`), and are instance-tagged so orphaned popups get cleaned (`:180`). C# `DisposeAsync` catches `JSDisconnectedException` and disposes `_selfRef` (`RichTextEditor.razor:480-493`). Init failures surface an inline error banner + `OnInitError` (`:266-274`). Word-import has a pre-allocation size guard (`:378`).

### 🟡 Problems / a11y gaps
- Floating bubble menu and drag handle are JS-rendered raw `<div>`/`<button>` with `aria-label` but **no `role="toolbar"`/roving focus and not keyboard-reachable** (`rich-text-editor.js:391-415`, `:486`) — selection formatting is mouse-only via the bubble. The static `EditorToolbar` is fine (`role=toolbar`, `aria-pressed`, `aria-label`, mousedown-preventDefault to keep selection — `EditorToolbar.razor:120-123`).
- Trigger/slash dropdown is keyboard-navigable (`role=option`/`aria-selected`) but its container has no `role=listbox`/`aria-activedescendant` wiring on the editor (`:179`,`:209`).
- `OnContentUpdate` round-trips full HTML on every edit (debounced JS-side) — fine, but no `MaxLength` enforcement is visible server-side (relies on JS `maxLength`).

### 🟢 Missing vs TipTap/Lexical
- No collaborative editing (Yjs/awareness), no markdown serialization/round-trip output (HTML only), no comments/suggestions/track-changes, no `@mention` provided by default (consumer supplies `Triggers`), no find-replace, no image-resize handles. Tables/task-lists/code-block/drag-handle/bubble/slash all present.

[...]

Issue #218, opened 2026-06-17
[audit] Drawer — no snap points, distance-only dismiss


Brain2k-0005:
**Component:** `Drawer` (`Lumeo`) · **vs** vaul (Drawer) / shadcn Drawer
_Audit 2026-06-17 · Lumeo 3.13.2 source · Radix + shadcn benchmark._

### 🔴 Bugs / correctness
- None verified. `RegisterDrawerSwipe` (`DrawerContent.razor:120`) matches the C# overload; focus-restore (`:111`/`:147`), nested-Escape (`:27`), transform-cleanup (`:116`) verified.
- (unverified) `mt-24` is baked into the base class for ALL sides (`DrawerContent.razor:92`) — only meaningful for a bottom drawer; on Top/Left/Right it fights `inset-y-0 h-full`. Worth a visual check.

### 🟡 Problems / a11y gaps
- Hardcoded backdrop `bg-black/80` (`DrawerContent.razor:23`) ignores `--color-overlay-backdrop`.
- No built-in close (X) button (unlike SheetContent); only swipe (touch) + Escape unless consumer adds `<DrawerClose>` — desktop discoverability gap.
- `div[role=button]` triggers; dangling aria IDREFs (same as Sheet).

### 🟢 Missing vs vaul
- No snap points (`snapPoints`/`activeSnapPoint`) — vaul's defining feature, entirely absent.
- No velocity/flick-based dismiss — JS uses fixed distance thresholds only (`components.js:812-813`).
- No background scaling (`shouldScaleBackground`).
- No nested-drawer recession; no `dismissible` vs `modal` split (collapsed into `PreventClose`, which also kills swipe at `:118`).
- No exit animation; no `asChild`.

### Severity: P1 — missing snap points + velocity dismiss + background-scale means Drawer is a bottom-Sheet, not a vaul Drawer (primary benchmark capability absent).

Brain2k-0005:
Deferred from the 3.18.0 backlog bundle (PR #344): this needs a dedicated effort rather than a rushed blind change — drag-gesture snap-point physics in the Drawer JS, which needs real-device verification. Tracked in #337.

Brain2k-0005:
Closing — the core gaps from the title are resolved in **3.19.0** (#345):

- vaul-style **snap points** — `SnapPoints` + two-way `ActiveSnapPoint`/`ActiveSnapPointChanged`
[...]

PR #345, opened 2026-06-18
feat: Drawer snap points + velocity dismiss (#218) · RichTextEditor a11y (#320)

Brain2k-0005:
Bundled feature work for the next release (**3.19.0**), targeting the two real P1 audit gaps. All additive and opt-in.

> ⚠️ **Needs browser verification before release.** Both features are gesture/JS-driven and cannot be runtime-verified in headless CI. Please verify on the Cloudflare preview (checklist below). The version bump / tag / NuGet publish happen **after** your OK (Rule #2).

## #218 — Drawer: vaul-style snap points + velocity dismiss
- **`SnapPoints`** (`IReadOnlyList<double>`, fractions `0<f≤1`, ascending) + **`ActiveSnapPoint`** / **`ActiveSnapPointChanged`** (two-way). A Top/Bottom drawer rests at fractional heights, drags between them, and dismisses when flicked/dragged below the lowest snap. Programmatically setting `ActiveSnapPoint` moves it.
- **Velocity / flick dismiss**: a fast flick in the dismiss direction closes even below the distance threshold (`LumeoGestureOptions.SwipeDismissVelocity`, default `0.4` px/ms; `0` = distance-only).
- **Backdrop** now uses the `--color-overlay-backdrop` token instead of a hardcoded `bg-black/80`, matching Sheet/Dialog.
- Implementation isolates the new snap gesture (`registerDrawerSnap`) so the existing, well-tested swipe path is untouched; in snap mode the JS owns the transform (CSS slide-in suppressed).

## #320 — RichTextEditor: a11y for the floating menus
- **Floating bubble toolbar is now keyboard-operable**: `Alt+F10` moves focus into it (ARIA Authoring-Practices pattern), `Arrow`/`Home`/`End` rove between buttons (roving tabindex), `Escape` returns focus to the editor. Stays open while focus is inside it.
- **Slash/mention listbox** wires `aria-activedescendant` (+ `aria-controls` / `aria-expanded`) on the editor textbox to the active option id, so screen readers announce the highlighted item as you arrow through. Cleared on close.

## Tests & build
- `dotnet test` — **3733 passing** (6 new Drawer-snap surface tests + the two backdrop tests updated for the token). Core + Editor build clean; both JS files pass `node --check`.
- bUnit covers the C# surface (params, snap-mode animation suppression, backdrop token, option default). The touch/keyboard behaviour is what the preview checklist below verifies.

## Verification checklist (Cloudflare preview)
**Drawer (#218)** — on a touch device / emulator:
- [ ] A drawer with `SnapPoints="[0.4, 0.75, 1]"` opens to the active snap and rests there.
- [ ] Dragging settles to the nearest snap; a fast flick down moves down a snap / dismisses from the lowest.
- [ ] A quick flick dismisses even with a short drag; a slow short drag snaps back.
- [ ] Backdrop tint matches Sheet/Dialog in both light and dark mode.

**RichTextEditor (#320)** — keyboard only:
- [ ] Select text → `Alt+F10` focuses the bubble toolbar; `←`/`→` move between buttons; `Esc` returns to the editor.
- [ ] Type `/` → arrow through the slash menu; a screen reader announces each highlighted item (aria-activedescendant).

## Deliberately out of scope (kept for dedicated efforts)
TipTap collaborative editing (Yjs) / markdown round-trip (#320 remainder); Drawer background-scaling & `dismissible`-vs-`modal` split (#218 remainder). These are larger and tracked under the audit index (#337).

---
_Generated by [Claude Code](https://claude.ai/code/session_01XPntycNyvABSR9U5ZKRBoU)_

## Release Notes

* **New Features**
  * Drawer snap points (touch): drawers can rest at multiple configurable heights, with two-way control via `ActiveSnapPoint`.
  * Velocity-based swipe dismissal: quick flicks can dismiss drawers in addition to distance-based swipes.
  * Editor selection (bubble) menu keyboard support, including focus return via keyboard.

* **Accessibility**
  * Improved screen-reader support for editor suggestion lists with correctly tracked active options.

* **Documentation / Demos**
  * Added a Drawer snap points (touch) demo and updated Drawer API docs.

* **Tests**
  * Expanded drawer snap-point and backdrop coverage.

### DIFF
--- a/docs/Lumeo.Docs/Pages/Components/DrawerPage.razor
+++ b/docs/Lumeo.Docs/Pages/Components/DrawerPage.razor
@@ -140,6 +140,41 @@
         </Drawer>
     </ComponentDemo>
 
+    <ComponentDemo Title="Snap points (touch)" Code="@_snapCode">
+        <Stack Gap="3">
+            <Lumeo.Text Size="sm" Color="muted">
+                Bottom drawer with snap points at 40% / 75% / 100%. On a touch device (or DevTools device emulation),
+                drag the handle to move between snaps; a quick flick down dismisses. On desktop it opens at the active
+                snap so you can still see the resting position. Active snap: <strong>@((int)(_snapPoints[_snapIndex] * 100))%</strong> (index @_snapIndex).
+            </Lumeo.Text>
+            <div>
+                <Drawer @bind-Open="_snapOpen">
+                    <DrawerTrigger>
+                        <Button Variant="Button.ButtonVariant.Outline">Open Snap Drawer</Button>
+                    </DrawerTrigger>
+                    <DrawerContent SnapPoints="_snapPoints" @bind-ActiveSnapPoint="_snapIndex">
+                        <DrawerHeader>
+                            <DrawerTitle>Snap Drawer</DrawerTitle>
+                            <DrawerDescription>Drag the handle to snap between 40%, 75%, and full height.</DrawerDescription>
+                        </DrawerHeader>
+                        <Stack Gap="2" Class="p-4">
+                            @foreach (var i in Enumerable.Range(1, 8))
+                            {
+                                <Flex Justify="between" Class="text-sm border-b border-border/40 pb-2">
+                                    <Lumeo.Text As="span" Size="sm" Color="muted">Row @i</Lumeo.Text>
+                                    <Lumeo.Text As="span" Size="sm" Weight="medium">Detail @i</Lumeo.Text>
+                                </Flex>
+                            }
+                        </Stack>
+                        <DrawerFooter>
+                            <DrawerClose><Button Class="w-full">Close</Button></DrawerClose>
+                        </DrawerFooter>
+                    </DrawerContent>
+                </Drawer>
+            </div>
+        </Stack>
+    </ComponentDemo>
+
     <ComponentDemo Title="Side — interactive" Code="@_sideInteractiveCode">
         <Stack Gap="4">
             <Lumeo.Text Size="sm" Color="muted">Pick a value to see it applied live.</Lumeo.Text>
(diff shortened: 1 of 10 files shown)


Brain2k-0005:
Addressed the CodeRabbit + Codex review in `b9a9f435`. Summary of what changed and the one finding I deliberately skipped:

**Fixed**
- **Snap honors `velocity: 0`** — an explicit `SwipeDismissVelocity = 0` now gives distance-only behavior in snap mode too (was forced back to `0.4`), matching the swipe path.
- **`PreventClose` + `SnapPoints`** — a protected drawer now still snaps between points (gesture registered with `dismissible: false`; dragging past the lowest snap settles back instead of closing) rather than opening fully and ignoring snap.
- **Vetoable snap dismiss** — the snap dismiss path now awaits a bool (`OnDrawerSnapDismiss`); if `OnBeforeClose` cancels, JS snaps back to the active snap instead of leaving the panel translated off-screen.
- **Gesture-mode tracking + mid-open re-registration** — the component records whether it registered snap or swipe, uses that for cleanup/programmatic updates, and re-registers if `SnapPoints`/`Side` change while the drawer is open (covers empty→populated `SnapPoints`).
- **Bubble toolbar focus trap** — `Tab` from the toolbar returns focus to the editor (keeps focus inside a modal trap rather than escaping to the body-level toolbar), and a `focusout` handler hides it once focus leaves entirely.
- **Editor listener cleanup** — `destroy()` now unregisters the `selectionUpdate`/`blur`/`focus` listeners via `editor.off(...)`.

**Skipped (with reason)** — CodeRabbit: *"snap gesture should honor `activationPx`/`firePx`"*. Snap mode intentionally tracks the finger immediately (a 60px activation deadzone — `SwipeDismissThresholdPx`'s value — would make snapping feel laggy), and the dismiss-past-lowest distance is size-relative (a fraction of the snap gap) so it adapts to drawer height, whereas an absolute `firePx` can exceed a short drawer's drag range. The one option that does apply to snapping — velocity — is honored (see the `velocity: 0` fix above). `VELOCITY_WINDOW_MS`/`DISMISS_FRACTION` have no corresponding `LumeoGestureOptions` entry, so there's nothing to thread through for them.

All 3734 unit tests pass; gesture/keyboard behavior is what the Cloudflare-preview checklist in the PR description verifies.

---

Issue #68, opened 2026-05-26
bug: Charts — Gauge div-by-zero, ResizeObserver leak on refresh, WordCloud extension race


Brain2k-0005:
From the deep static bug-scan of `src/Lumeo.Charts/**`. Most chart wrappers (30+ chart types) are clean — these are the residual defects.

## HIGH — Gauge division-by-zero
`src/Lumeo.Charts/UI/Chart/GaugeChart.razor:75`

`(Max - Min)` range normalisation throws `DivideByZeroException` when `Min == Max`. Also: if `Ranges` is empty but accessed via `.Max(r => r.Threshold)` the sequence is empty and `InvalidOperationException` is thrown.

```csharp
// current
var t = (Value - Min) / (Max - Min);   // ← Min == Max → div-by-zero

// fix sketch
var range = Max - Min;
var t = range == 0 ? 0 : (Value - Min) / range;
```

## HIGH — ResizeObserver leak on chart refresh
`src/Lumeo.Charts/wwwroot/echart.js:432` (or near) — `refreshAllCharts()` disposes the old chart instance but does NOT call `_lumeoObserver.disconnect()` before disposal. After refresh, the observer is orphaned and continues firing into a destroyed ECharts instance. Net effect: memory leak across theme switches / data refreshes.

```js
// before dispose
if (entry.observer) {
  entry.observer.disconnect();
  entry.observer = null;
}
chart.dispose();
```

## MED — WordCloud `Task.Delay(200)` race
`WordCloudChart.razor:64` — after `loadExtension('wordcloud')` a hardcoded `Task.Delay(200)` waits for the extension to register. On slow / cached CDNs the extension may not be ready in 200 ms (race), or may be ready in 5 ms (waste). Fix: have the JS side resolve a Promise on `define()` complete and `await` it from .NET.

## MED — Full `setOption` per parameter change
Every chart's `OnParametersSet` rebuilds the whole `_option` object and calls ECharts `setOption(option, true)` (full replace). For streaming-data scenarios (live dashboards) this thrashes ECharts. Fix: diff the option (or expose an `appendData` API on each chart) so consumers can stream data without full re-render. Large but worth scheduling.

## LOW
- Custom `Func<ChartDataContext, string>` formatter not supported for `LabelFormat` — tooltips / labels limited to ECharts' built-in `{b}`, `{d}`, `{v}` substitutions. Add a formatter parameter.
[...]

PR #347, opened 2026-06-24
feat: Radix/Base-UI/shadcn parity audit — a11y, RTL, OKLCH, FormGenerator, MCP/CLI

Brain2k-0005:
## Summary

A Radix / Base-UI / shadcn **parity audit pass** across the whole project — accessibility, RTL, theming, the LumeoFormGenerator, and the MCP/CLI tooling. Everything is **additive and opt-in**; existing usage is unchanged (the OKLCH and logical-utility changes are visually/behaviourally identical in LTR — verified pixel-for-pixel in the browser).

See the `[Unreleased]` section in `CHANGELOG.md` for the full list. Version bump intentionally left out.

## What's in it

**Accessibility**
- FormField `aria-describedby` now reaches every form control (Checkbox, Switch, RadioGroup, Slider, Select, Combobox, …), not just Input; single-focus controls also adopt the field `ControlId` so `<label for>` resolves.
- `aria-current` on the active Sidebar nav item + Scrollspy link; overlay entry animations honour `prefers-reduced-motion`; Cascader arrow-key roving; Menubar trigger/item roles; Avatar status no longer colour-only; Chart `AriaLabel`; asChild on AlertDialog/Drawer triggers.
- Overlay scroll-lock compensates for scrollbar width (no layout shift).

**RTL (new)**
- Migrated the component library's directional Tailwind utilities to **logical** ones (`ml→ms`, `left→start`, `text-left→text-start`, `rounded-l→rounded-s`, `border-l→border-s`, …) — LTR-identical, mirrored under RTL.
- New `DirectionProvider` component (sets native `dir`). Verified in-browser: under `dir=rtl` the Alert icons move right + text right-aligns; LTR untouched.

**Theme**
- The entire colour palette (base + all 8 themes, **878 tokens**) migrated HSL → **OKLCH** (exact 1:1, brand unchanged), matching Tailwind v4 / current shadcn.

**LumeoFormGenerator**
- New mappings (TimeOnly/TimeSpan→TimePicker, `List<string>`→TagInput, MultilineText→Textarea, Phone/Url→typed Input), validation→attribute mapping (`[Range]`→Min/Max, `[StringLength]`→MaxLength+counter), `bool` not implicitly required, nullable-numeric clear-to-null, `[Display(Order)]` sorting. Real compile-the-output test.

**MCP / CLI**
- Type-bound enum validation (`Size="Large"` is now caught), cascading-gated parent-child rule, per-component test-coverage + `[EditorRequired]` surfaced, new `lumeo_get_a11y` tool, README refreshed.

**New component features**
- Tabs `IconReveal` (animated icon→label reveal), Card `CardTitle`/`CardDescription`, public `Lumeo.Cx`.

## Docs

- New DirectionProvider page; **179 new demos across 107 component pages** filling previously-undemonstrated major features (audited every page against the real component API); CHANGELOG + SECURITY brought current.

## Testing

- **4029** bUnit + **31** source-generator + **15** MCP tests green; full solution + docs build clean under `-warnaserror`. RTL/OKLCH/Tabs visually verified in the browser.

## Review notes

Good candidates for a close look: the logical-utility RTL migration (broad but mechanical + LTR-safe), the OKLCH conversion, and the docs demos.


Brain2k-0005:
@codex review

Thanks for the earlier review (commit `3441f57`). All 8 findings were verified as real — **6 are fixed** in this branch, 2 deferred to dedicated follow-up PRs (each needs its own test coverage and neither is a regression from this PR).

**Fixed**
- **FormField labels** — `Textarea` & `DatePicker` now adopt the FormField `ControlId` via an `EffectiveId` (auto-resize / focus interop switched to match), so `<label for>` points at the rendered element.
- **RTL arrows** — `Popover`/`Tooltip`/`HoverCard` arrow side-offsets + outward borders reverted to physical (incl. the default cases) to match the physical JS `positionFixed` placement.
- **DirectionProvider** — `Carousel`, `ResizableHandle`, `SplitterDivider`, `Stepper`, `TabsList`, `TabsTrigger` now consume the scoped `LumeoDirection` cascade via `EffectiveDirection` (falls back to the theme direction).
- **CLI add-fail** — propagates a failed `dotnet add package` (`Environment.ExitCode = 1`) when there's a consumer project; skips gracefully with a hint when there's none (so bare-dir `--vendor` still succeeds).
- **CLI preview** — `--diff`/`--view` no longer runs package installs (gated on `writeAllowed`).
- **CLI satellite deps** — a non-`--vendor` satellite no longer vendors its source dependency graph (NuGet pulls them transitively).

**Deferred (follow-up PRs)**
- `lumeo_validate_markup` nesting is set-membership, not a tag stack — needs a stack-walk rewrite + regression tests.
- Satellite raw-URL 404 — needs the CLI `DeriveFileBaseUrl` change and `deploy-cloudflare.yml` publishing to land in lockstep, plus an HTML-shell guard.

CI / E2E / docs / mcp-server / CodeRabbit all green.

Brain2k-0005:
@codex review

Updates since your last review (`254751e7`), all CI + E2E green:

**Your round-2 findings — fixed**
- **DropdownMenu** RTL arrow: swapped the logical side-offsets/borders to physical to match the physical JS placement (incl. the default case), same as Popover/Tooltip/HoverCard.
- **TagInput** / **TimePicker**: the inner text input / trigger now adopt the FormField `ControlId` via `EffectiveId` (focus + anchor interop switched to match), so `<Label For>` resolves to the real focusable element.
- **tweakcn adapter self-reference** (your catch): the opt-in `lumeo-tweakcn.css` had a cyclic `var(--primary, var(--color-primary))` fallback. It's **removed** and replaced with native indirection — see below.

**New: native 1:1 shadcn/tweakcn theming**
All 886 `--color-*` token definitions (base `:root`/`.dark`, the `[data-base-color]` ramps, and the 8 palette files) are now `--color-X: var(--X, <lumeo-default>)`. A bare shadcn `--primary` drives Lumeo with zero setup; defaults are preserved when unset (verified in-browser). `--radius` is already 1:1. `var(--color-X)` usages untouched.

**Docs/UX fixes**
- Confetti burst is `position:fixed` so it no longer adds a page scrollbar; idle canvas is zero-size and the trigger stays centred.
- AI demo pages are fully reflow-proof (every auto-growing/collapsible/streaming demo is in a fixed-height box; no `min-h` on dynamic demos).
- New `NumberInput` `Block` (full-width) param; the LumeoForm generator emits it so generated number fields line up with sibling inputs.

Still intentionally deferred to follow-up PRs (not regressions from this branch): `lumeo_validate_markup` nesting (needs a tag-stack rewrite + tests) and the satellite raw-URL 404 (CLI + deploy-workflow must land in lockstep).

Brain2k-0005:
@codex review

Thanks for the round-2 review (commit `18c22562`). All 7 P2/P3 findings are addressed at HEAD (`19086ebe`), and an adversarial verification pass also swept the codebase for sibling instances of each bug class:

- **DataGridCell OKLCH**: `hsl(var(--primary)/.05)` → `color-mix(in oklab, var(--color-primary) 5%, transparent)` so it survives a shadcn/tweakcn OKLCH `--primary`.
- **Cascader search keystrokes**: mirrored TreeSelect's `OnSearchKeyDown` forwarder (caret keys stay in the box; Up/Down/Enter/Escape forward) instead of a blunt stopPropagation that swallowed Escape-to-close.
- **DateTimePicker disposal**: `IAsyncDisposable` + broadened catch (ObjectDisposed/InvalidOperation); TimePicker too.
- **Drawer/Sheet RTL swipe**: `SwipeDirection` now derives from an RTL-aware `EffectiveSide` (position is logical + the slide keyframes mirror under `[dir=rtl]`).
- **FormField `<Label for>`**: ControlId is now adopted by Select **and the siblings the sweep found** — Cascader, TreeSelect, Combobox, InputMask, Mention, Rating.
- **Select keyboard nav**: `RegisterItem` upserts + `SelectItem` re-registers on change, so an item disabled while the menu is open leaves the nav set.
- **DataGrid RTL**: grouping clear-all `ml-1`→`ms-1`, plus ~15 more physical→logical utilities across the DataGrid family (Toolbar, FilterPill, Footer, GroupRow, toolbar dropdown buttons, ColumnVisibility, dirty-cell marker, `ml-auto` panels).
- Also from the sweep: `components.js` Calendar/Gallery/Tabs swipe emitters made RTL-aware; SelectContent Home/End search-guard; DataGridColumnFilter stop-propagation; MegaMenuItem re-reports Disabled.

New feature since your review: `MapMarker.Highlighted` (`19086ebe`) — `clusterProperties` aggregation + a data-driven cluster `circle-color` tinted via `--lumeo-map-cluster-highlight`.

Full bUnit suite green (4030); CI / E2E / docs-build all green at HEAD.

Brain2k-0005:
@codex review

Another pass would be appreciated — a batch of doc-accuracy + CI-hardening changes landed since your last review, and all checks are green now. Highlights to focus on:

**Docs-accuracy pass (4.0.0 "no fake data")**
- `/docs/bundle-facts`: replaced the estimated competitor sizes with **real measured** brotli-q11 numbers of each core DLL from its latest-stable public .nupkg (FluentUI 302 / Lumeo 627 / Radzen 1,003 / MudBlazor 1,318 / Telerik 1,480 KB). Sorted ascending (neutral), and the footnote states the exact methodology + what's excluded (IL trimming, JS/CSS assets — including Lumeo's own). Deliberately does **not** claim Lumeo is smallest (FluentUI is). Please sanity-check the wording so it can't read as an unfair/indefensible claim.
- `/docs/changelog` + root `CHANGELOG.md`: added the 4.0.0 entry, including the NuGet-free standalone eject.
- `/docs/cli`: documented `lumeo eject`, `init --standalone`, `add --vendor`, `deps install`; rewrote the old "runtime services can't be vendored" section into a default-vs-standalone explanation; `all five satellites` → all ten; fixed the registry URL (`…/registry/registry.json` → `…/registry.json`).
- README / SECURITY.md / MIGRATION.md / `skills/lumeo`: brought current for 4.0.0 (164 components, 13 MCP tools, supported-versions table, `lumeo.config.json` → `lumeo.json`).

**Bug fix**
- Header logo had `Href=""` (no-op) → `Href="/"` so the wordmark returns to the landing page.

**CI hardening** (build-and-test had been red on a three-layer cascade)
- `-warnaserror`: 45 xUnit1031 blocking-task sites → `await` (39 methods → `async Task`), 2 xUnit2031 `Assert.Single(Where(..))` rewrites, removed QRCode's dead `_margin` field.
- Calendar / DateRangePicker flicker tests asserted a raw `"March 2024"` substring that matched an adjacent-month day-cell aria-label under CI's invariant culture (`dd MMMM yyyy`); now scoped to the month-panel `aria-label="MMMM yyyy"`.
- Regenerated the per-component registry (qr-code / result) for the source edits.

The standalone-eject docs and the bundle-facts comparison are the two areas I'd most value your eyes on.

Brain2k-0005:
@codex review

Thanks for the last pass — all 10 findings were real and are now fixed (commits `eca1d178` CLI + `85543c72` docs). Mapping so you can verify:

**P1**
- **Standalone `@using Lumeo` placement** — now written to the **project-root** `_Imports.razor` (not under `componentsPath`), so it cascades to app pages too. Standalone init e2e test updated to assert the root location.
- **`_content/Lumeo` JS 404 in standalone** — `EnsureRuntimeVendoredAsync` now also mirrors the core JS/CSS to `wwwroot/_content/Lumeo/` (where the verbatim-vendored interop loads `_content/Lumeo/js/components.js`, etc.).
- **Satellite `--vendor` raw-URL 404** — `DeriveFileBaseUrl` now nests `/raw/<package>/` for the self-hosted layout, and `deploy-cloudflare.yml` publishes each `src/Lumeo.<Sat>/` to `release/wwwroot/raw/<package>/` so the two move in lockstep.

**P2**
- **bundle-facts internal contradiction** — re-measured the real 4.0.0 `.nupkg` (~1.09 MB; DLL 878 KB zip / 627 KB brotli; PDB now ships separately as `.snupkg`). The breakdown + all version/date refs are updated; nupkg 1.09 MB > DLL 878 KB > brotli 627 KB, so the "package smaller than its DLL" contradiction is gone.
- **deps install airgap overclaim** — qualified: ESM-only deps (FullCalendar/CodeMirror) are skipped and still load from their CDN.
- **eject regex** — now strips the expanded `<PackageReference>…</PackageReference>` form too, not only self-closing.
- **eject continues after runtime-vendor failure** — `EnsureRuntimeVendoredAsync` returns a success bool; `eject` now aborts and leaves the project untouched (no half-ejected state); `add` honours it too.
- **DataGrid export backend on `--vendor`** — emits a note that Excel/PDF export needs the compiled `Lumeo.DataGrid.Export` backend (can't be source-vendored).
- **deps install `--lib` bootstrap** — now built from the **downloaded** (filtered) deps, not the whole manifest, so un-downloaded keys (Map/PdfViewer) aren't rewritten to `/lib/lumeo-vendor/`.
- **catalog `?category=` query-only nav** — now re-syncs on `NavigationManager.LocationChanged` (unsubscribed in `Dispose`).

Verified locally: full solution `-c Release -warnaserror` clean, Cli.Tests 43/43 (incl. 7 standalone build proofs), registry-freshness gate clean. A fresh look at the standalone-asset wiring and the satellite raw-URL/deploy pairing would be especially welcome.

Brain2k-0005:
@codex review

Round 2 — thanks, all 9 follow-ups were real and are now fixed (commit `afb8008d`):

**Standalone / eject completeness**
- **Core JS mirror completed** — `s_standaloneCoreAssets` now mirrors every core runtime module (adds `signature-pad.js` + `toolbar.js` alongside `components.js`/`theme.js` + both CSS) under `wwwroot/_content/Lumeo/`, so Toolbar/SignaturePad no longer 404.
- **Mandatory-asset failure propagates** — a failed core-asset fetch now returns `false` from `EnsureRuntimeVendoredAsync`, so eject won't strip the package over a runtime that will 404.
- **eject writes the `_Imports` bridge** — extracted into a shared `EnsureStandaloneImportsAsync` used by both `init --standalone` and `eject`.
- **`_Imports` merge is per-line** — appends only the individually-missing `@using` lines (whole-line match), so an existing `@using Lumeo.Components` no longer masks `@using Blazicons` / `Lumeo.Services` / etc.
- **eject aborts on a failed component re-vendor** — checks each `Add` result and leaves the package in place if a stale `lumeo.json` key fails, instead of stripping over missing source.

**Other**
- **DataGrid `--vendor` note** now points at a package that exists — the export backend ships inside the `Lumeo.DataGrid` NuGet (`Lumeo.DataGrid.Export` is `IsPackable=false`), so it tells users to reference `Lumeo.DataGrid`, not a non-existent package.
- **`lumeo update` refreshes vendored satellite assets** — re-copies `wwwroot/_content/<package>/…` JS/CSS so updated Razor/C# isn't left calling stale interop.
- **FileUpload dispose race** — cancels a dispose `CancellationTokenSource` first; `WaitAsync` observes it and returns before the `Release` finally; the semaphore is left to GC (no `ObjectDisposedException` / lost release).
- **Catalog `?category=`** — the `LocationChanged` handler now clears filters when the category is absent/unknown, so navigating back to `/components` doesn't stay stuck.

Verified locally: full solution `-c Release -warnaserror` clean; Lumeo.Tests 4831, Cli.Tests 43 (incl. 7 standalone e2e), Docs 28, RegistryGen 32, SourceGen 31; registry regenerated for the FileUpload source change.

Brain2k-0005:
@codex review

(Re-triggering — the previous request hit a usage limit.) All review feedback through round 2 is addressed on the current branch head `afb8008d`. The detailed change list is in the comment above; in short:

- **Round 1** (10 findings): standalone `_Imports` placement, `_content/Lumeo` JS mirror, satellite `--vendor` raw-URL + deploy, eject regex/robustness, DataGrid-export note, `deps install --lib` bootstrap scope, bundle-facts re-measure, airgap wording, catalog query-nav.
- **Round 2** (9 follow-ups): completed the core-JS mirror (toolbar/signature-pad), mandatory-asset failure now aborts the vendor, `eject` writes the `_Imports` bridge + appends only missing usings + aborts on a failed component re-vendor, DataGrid note points at the real `Lumeo.DataGrid` package, `lumeo update` refreshes vendored satellite assets, FileUpload dispose race, catalog filter-clear on empty category.

Verified locally: full `-c Release -warnaserror` clean; Lumeo.Tests 4831, Cli.Tests 43, Docs 28, RegistryGen 32, SourceGen 31 green; registry regenerated. A fresh pass on the standalone/eject runtime wiring would be appreciated.

Brain2k-0005:
@codex review

Round 3 — all 7 addressed (commit `a49b81b0`):

**P1 — eject strips unvendored satellites:** eject now strips only core + the satellite package of each *re-vendored* component. A satellite used via the default NuGet flow (no `--vendor`) has no vendored source, so its package is **kept** (with a note to `add … --vendor` it first) rather than removed — no more "neither package nor source".

**P2**
- **standalone persisted before abort:** `standalone:true` is now saved only after all abortable work succeeds (right before stripping), so a mid-loop abort no longer leaves a half-ejected config that retries skip as "already standalone".
- **runtime sentinel too coarse:** runtime vendoring is now self-healing — it (re)copies each *missing* file instead of trusting `Cx.cs` as proof of completeness.
- **satellite-asset failure swallowed:** a failed satellite asset fetch now sets a non-zero exit (eject aborts; `add --vendor` fails) instead of a cosmetic "skipped".

**P2/P3 cross-cutting**
- **FileUpload wrapper:** now variant-aware — Button/Avatar stay inline, Dropzone/default stretch to width.
- **RegistryGen type-dep scan:** strips comments first, so a doc-only `<see cref="TabsLayout"/>` no longer makes `add resizable` vendor Tabs (resizable/speed-dial lose the comment-only deps).
- **validate_markup nesting:** parent-component check is now an open-element-stack walk, so `<DialogTrigger/> … <Dialog/>` siblings fail instead of passing flat presence (+3 regression tests, 18/18).

Verified locally: full `-c Release -warnaserror` clean; Lumeo.Tests 4831, Cli.Tests 43, Docs 28, RegistryGen 32, SourceGen 31, MCP 18/18; registry regenerated.

Brain2k-0005:
@codex review

Round 4 — all 5 addressed (commit `fc6622e9`):

**CLI eject edges (mine)**
- **`add --diff`/`--view` no longer writes the runtime/assets** in standalone preview — gated on the same preview condition as component files.
- **eject strip is attribute-order-independent** — `<PackageReference Version="…" Include="Lumeo" />` (or a leading `Condition=`) is stripped too, not just Include-first.

**Component bugs (pre-existing, surfaced by the review)**
- **ConsentBanner RTL:** centered banner switched from logical `start-1/2` to physical `left-1/2` (+`-translate-x-1/2`) so it stays centered in both LTR and RTL.
- **FormField/FormMessage aria:** with `AutoRenderMessage="false"`, `_errorId` is propagated through `FormFieldContext` so the child `FormMessage` emits `id="@ErrorId"` — `aria-describedby` now points at a rendered element (and the error is still announced).
- **Input controlled rollback:** track `_lastPushed`; in the controlled case (`ValueChanged.HasDelegate`) adopt any incoming `Value` that isn't the echo of our own edit, so a rejected/normalized value rolls the display back while in-flight typing is preserved.

Verified locally: full `-c Release -warnaserror` clean; Lumeo.Tests 4831 (Input 28, FormField/FormMessage 12), Cli.Tests 43, Docs 28, RegistryGen 32, SourceGen 31, MCP 18/18; registry regenerated.

Brain2k-0005:
@codex review

Round 5 — all 9 addressed (commit `72f618c8`):

**CLI eject**
- **CRITICAL regression fixed:** eject deferred persisting `standalone:true` until after the re-vendor loop, but each `Add()` reloads lumeo.json from disk and saw `false`, namespace-rewriting the re-vendored source out of the Lumeo namespace. Now persist standalone BEFORE the loop and roll it back to false on any abort.
- **Runtime refresh on upgrade:** standalone runtime is re-copied when the recorded version != the registry version (not only self-healed for missing files); `lumeo update` refreshes it too.
- **Single-quoted `Include='Lumeo'`** is now stripped by eject as well.

**RegistryGen**
- The markup `<Tag>` scan strips Razor/HTML/block comments first (line `//` left intact to avoid eating `https://` URLs), so a `<Tag>` mentioned only in a comment example no longer becomes a dep (Tooltip→Button etc. removed across many components).

**Components**
- **CodeEditor / RichTextEditor / SignaturePad:** the Input controlled-rollback fix applied library-wide — track `_lastPushed`, and in the controlled case re-sync the editor when the incoming Value isn't the echo of our own edit, so a rejected/normalized value rolls the display back.
- **RichTextEditor:** also pushes aria-invalid/aria-describedby to the live contenteditable after validation changes (new additive `RichTextSetAriaAttributesAsync` interop, default no-op).
- **Input:** inside a FormField, a consumer-splatted `id` no longer overrides the generated control id, so `<label for>` == `<input id>`.

Verified locally: full `-c Release -warnaserror` clean; Lumeo.Tests 4840 (+9 regression tests), Cli.Tests 43, Docs 28, RegistryGen 32, SourceGen 31, MCP 18/18; registry regenerated.

Brain2k-0005:
@codex review

Round-6 follow-up pushed (e1d90705): library-wide sweep of the two systematic patterns you flagged.

- **Controlled-component rollback** (`_lastPushed` discriminator, branch on `*Changed.HasDelegate`): Switch, ToggleGroup, DropdownMenuCheckboxItem, Checkbox, DatePicker — a controlled parent veto now rolls the visible state back; an echo of our own push is ignored; the uncontrolled same-value guard is preserved.
- **FormField splat-id** (`EffectiveAttributes` strips a consumer `id` when `InsideFormField`; `EffectiveId` owns the generated `FormFieldControlId`): PasswordInput, Switch, InputMask — the `<label for>` association no longer breaks when a consumer splats an `id`.
- `IComponentInteropService.SaveFocus/RestoreFocus` now have default no-op impls (additive; real impl unchanged).

New rollback + FormField tests per component; full suite 4867 green; Release `-warnaserror` clean; registry regenerated.

Brain2k-0005:
@codex review

Round-7 follow-up pushed (659e65fd) — all 6 P2 findings from your last review addressed:

1. **Checkbox** — inside a FormField the generated `ControlId` now owns the `<button id>` (a consumer `Id` no longer wins over it); `Id` still owns a standalone checkbox.
2. **Textarea** — strips a consumer-splatted `id` when `InsideFormField` (`EffectiveAttributes`), matching Switch/PasswordInput/Checkbox.
3. **DatePicker** — `HandleDateSelected` / preset / clear now refresh `_inputBuffer` on pick, so a controlled echo no longer leaves stale draft text (regression test proven to fail without the fix).
4. **TimePicker** — `RovingValue` no longer seeds a disabled selection as the tab stop; falls back to the first enabled option.
5. **DateTimePicker** — added a forward-registration branch so enabling `ShowSeconds` while the popover is open registers the seconds column's nav keys.
6. **CLI** — `init --standalone` no longer writes root `_Imports` before the runtime is vendored; the import-write moved into `add` (post-vendor) and `eject`.

New regression tests for #1–#4; full suite 4873 green; Release `-warnaserror` clean; registry regenerated.

Brain2k-0005:
@codex review

Round-8 pushed (6dcd2d93) — all 3 findings addressed, plus a comprehensive sweep of the splat-id pattern's tail so it stops recurring:

- **FormField splat-id, library-wide**: audited all 22 FormField-owned-root controls. 6 were genuinely affected and fixed via the established `EffectiveAttributes` (strip consumer `id` when InsideFormField): **DateTimePicker, RadioGroup, Rating, TimePicker** (direct), and the two picker-triggers you named — **SelectTrigger, ComboboxInput** (the FormField id reaches them via the parent context, so the parent now passes an `InsideFormField` flag through SelectContext/ComboboxContext). The other 16 are already-fixed or not-affected (splat sits on a wrapper div with a self-generated id, not on the FormField-owned element).
- **Eject Blazicons (P2)**: `EnsureRuntimeVendoredAsync` now ensures `Blazicons.Lucide` whenever the standalone runtime is vendored (idempotent), so a first `add separator/center/text` compiles.
- **DataGrid group chip (P3)**: chip wrapper is now a `<div>` (inline-flex) instead of a `<span>`, so the Tooltip's block `<div>` is a legal child and the remove button stays inside the chip.

Regression tests for all 6 newly-fixed controls; full suite 4885 green; Release `-warnaserror` clean; CLI tests 43 green.

Brain2k-0005:
## Codex review loop — closing summary

The two **systematic** patterns Codex surfaced are now closed library-wide:

- **Controlled-component rollback** — Switch, ToggleGroup, DropdownMenuCheckboxItem, Checkbox, DatePicker.
- **FormField splat-id** — audited all 22 FormField-owned-root controls; the 11 genuinely affected are fixed (Checkbox, Textarea, Switch, PasswordInput, InputMask, DateTimePicker, RadioGroup, Rating, TimePicker, SelectTrigger, ComboboxInput). The latest review surfaced **no** new splat-id issues — pattern drained.

**Eject / standalone** hardened end-to-end, including the Blazicons runtime dependency and aborting the vendor when that mandatory install fails.

Every fix landed with a regression test; full suite **4885 green**, Release `-warnaserror` clean, CLI tests 43 green, build-and-test + bom-check + docs CI green (the one E2E failure was the known-flaky splitter-drag smoke, rerun).

### Deferred (pre-existing, unrelated to the swept patterns)
Later rounds began surfacing the library's general bug backlog — genuine but pre-existing issues in unrelated areas, recorded here for a future focused pass rather than this campaign:

1. **P2** `Lumeo.SourceGenerators/LumeoFormGenerator.cs:891` — generated enum `Select` renders a blank trigger (no child content emitted).
2. **P2** `Tooltip/TooltipContent.razor:75` — Top/Bottom arrow off-center in RTL (`start-1/2` + physical translate).
3. **P2** `Tooltip/TooltipContent.razor:14` (+ components.js) — arrow stays on the wrong edge after a JS collision-flip (`dataset.side` not updated).
4. **P2** `Lumeo.DataGrid/.../DataGrid.razor:720` — Space-suppression cancels Space-activation of buttons inside a `CellTemplate`.
5. **P3** `DateTimePicker.razor:600` — Enter/Space on a time option double-fires selection (onkeydown + synthesized native click).

Thanks @codex — concluding the per-round loop here.

Brain2k-0005:
@codex review

Round-9 pushed (7fe07d03) — 4 of the 5 findings fixed:

- **Source-gen enum Select**: the generated `SelectTrigger` now emits a `switch` mapping the live enum value to the same humanized label its `SelectItem` shows (no more blank closed trigger).
- **Tooltip RTL arrow**: Top/Bottom arrow now uses the physical `left-1/2` (not logical `start-1/2`) so it stays centered under RTL.
- **Space/Enter button activation**: the shared `registerPreventDefaultKeys` `skipEditable` exemption now also covers `button, a[href], [role=button]`, so a container Space rule (DataGrid/FileManager/AudioPlayer) no longer cancels native activation of a button inside a CellTemplate/item.
- **TimePicker + DateTimePicker double-select**: Enter/Space are no longer handled in the column keydown path (native button click is the sole selection path), so activation fires `ValueChanged` once, not twice.

**Still open (next round):** the tooltip **collision-flip arrow** (data-side not updated when JS flips) — that needs a careful change to the shared `positionFixed` interop (returns the resolved side), which I'm doing as an isolated step with full overlay regression coverage.

New source-gen + double-select regression tests; full suite 4885 green, Release `-warnaserror` clean.

Brain2k-0005:
@codex review

Pushed 3b4e364f + fbfd6ad5 — round-9 finished and the round-10 RTL pattern swept:

- **Tooltip collision-flip arrow** (round-10 #1): `positionFixed` now RETURNS the resolved side; TooltipContent renders the arrow + data-side from it, so the arrow follows a flip. (Also fixed a registry-gate drift: components-api.json had been wrongly reverted as 'churn' — it tracks API surface + handled-key lists, restored now.)
- **RTL centering, library-wide** (round-10 #2/#3 + the pattern): swept every horizontally-centered overlay/arrow that paired logical `start-1/2` with the physical `-translate-x-1/2` → now physical `left-1/2`. Covers Popover / HoverCard / DropdownMenu arrows, Toast TopCenter/BottomCenter, Carousel nav, BottomNav indicator, Image/ImageGallery controls. Left/right anchors that SHOULD flip in RTL (e.g. ToastPosition TopLeft's `start-4`) left logical.

New flip + RTL-centering regression tests; full suite 4886 green, Release `-warnaserror` clean, registry regenerated.

Brain2k-0005:
@codex review

Pushed 9a7aeee6 — round-11, 3 of 4:

- **Sidebar off-canvas (RTL)**: the collapsed translate is now direction-aware (`-translate-x-full rtl:translate-x-full` for the start-0 Left panel, mirror for Right), so it slides off whichever physical edge its logical anchor resolves to.
- **AudioPlayer scrub fill (RTL)**: fill uses physical `left-0` to match the physical `OffsetX` seek math (no more mirrored fill-vs-seek in RTL).
- **RestoreFocus**: now restores only when focus was lost to `<body>` or still inside the dismissed surface — a controlled/background close no longer steals focus back to the old trigger.

**Still open (next, dedicated):** the tooltip arrow not re-syncing after a LIVE scroll/resize collision flip — it needs a JS→.NET side-change callback on the shared positionFixed (or a [data-side] CSS-driven arrow across all arrowed overlays), so I'm doing it as its own focused change rather than bundling.

New Sidebar + AudioPlayer regression tests; full suite 4888 green, Release `-warnaserror` clean.

Brain2k-0005:
@codex review

Pushed 35b1f65e — round-12, 3 of 4:

- **Arrow-key suppression on inner buttons** (regression of the round-9 button exemption): the button/link skipEditable exemption is now limited to ACTIVATION keys (Space/Enter); Arrow/Home/End/Page are suppressed on buttons again (AudioPlayer's Arrow seek no longer also scrolls the page).
- **Satellite add silent no-op** (follow-up to the round-9 bool return): the non-vendor path now verifies the satellite package is actually referenced before skipping the file copy; otherwise it errors + non-zero exits instead of printing OK with 0 files.
- **Window resize handle (RTL)**: stays physical `right-0` to match the physical resize math + cursor-se-resize glyph.

**Still deferred, each its own focused change next:** (a) the tooltip arrow after a LIVE scroll/resize collision flip — needs a JS→.NET side callback on the shared positionFixed; (b) FormField `<label for>` when a picker variant (TimePicker Wheel / DatePicker Inline) renders a child that never applies the cascaded ControlId.

New Window regression test; full suite 4890 green, Release `-warnaserror` clean.

Brain2k-0005:
@codex review

Pushed 971149b3 — round-13 (both findings were follow-ups on the round-12 satellite check):

- **Preview-safe**: the satellite package-presence check is now gated behind `writeAllowed`, so `add chart --dry-run` / `--diff` / `--view` stay non-mutating previews (show the planned install) instead of erroring.
- **No misleading OK**: a missing required satellite package now `return`s (aborts) with a clear error instead of setting ExitCode=1 then falling through to the "OK … 0 files written" summary.

Also fixed a registry-gate drift (a regression test was added after RegistryGen ran; regenerated as the final step).

**Next, with focus:** the two deferred items — tooltip arrow after a LIVE scroll/resize flip (positionFixed side callback), and FormField `<label for>` for picker variants that don't apply the cascaded ControlId (TimePicker Wheel / DatePicker Inline).

Brain2k-0005:
@codex review

Pushed 1ce464a1 + 080b3b08 — both remaining deferred items resolved (better than the original fixes):

- **FormField id wiring (supersedes the round-12-4 suppression)**: rather than dropping the label's `for` for TimePicker Wheel / DatePicker Inline / DatePicker Inline+Wheel, the FormField's real `EffectiveTriggerId` + `aria-describedby` + `role="group"` now splat onto TimeWheelPicker / Calendar / DateWheelPicker's own root via AdditionalAttributes — full label/help/error association on every variant, not just the popover-trigger ones.
- **Tooltip live collision-flip**: `positionFixed` now reports LATER side changes too (a scroll/resize reposition while the tooltip stays open), not just the initial placement — via a JS→.NET callback reusing ComponentInteropService's existing self-ref/dispatch pattern (a default-interface-method overload, so every other PositionFixed caller/implementer is untouched).

**In progress now:** a library-wide sweep for the controlled-rollback pattern your round-14 review found 3 fresh instances of (Dialog, DropdownMenuRadioGroup, TimeWheelPicker) — auditing all ~54 components with a bindable Parameter+EventCallback pair for the same anti-pattern, not just patching the 3 named ones.

Brain2k-0005:
@codex review

Pushed d17028c4 — library-wide controlled-rollback sweep (round-14's 3 named instances + 13 more found by auditing all ~54 bindable-Parameter components): Gantt, Scheduler, Cascader, DateTimePicker, Dialog, DropdownMenuRadioGroup, FileManager, HoverCard, InputMask, PickList, Scrollspy, Segmented, SortableList, TimeWheelPicker, TreeView, Window. Each has a regression test proving a controlled veto rolls back. Also fixed the missing satellite vulnerability-scan coverage in ci.yml (round-15 finding).

Two corrections found during central verification (not from your review, caught by re-running the full suite after the sweep):
- SortableList: the generic recipe conflicted with the pre-existing #144 contract (a fire-and-forget ItemsChanged observer must not have its local reorder reverted by an unrelated re-render) — reconciled so both the established #144 behavior AND genuine, distinguishable vetoes work correctly.
- A pre-existing (not sweep-related) date-fragile test in DateTimePickerRovingSeedTests was hardcoded to June 2026 and broke once "today" rolled past it — anchored to the current month instead.

Full suite 4952 green, Release -warnaserror clean.

Still working through your other round-15 findings (TimePicker/DatePicker ARIA labeling for group pickers, duplicate FormField ids in compound Wheel+ShowTimePicker DatePicker, Dialog description registry leak) as a follow-up push.

Brain2k-0005:
@codex review

Pushed 2a320ae9 — all 6 round-16 findings addressed:

- **PickList / FileManager / TreeView** (the same conflict already resolved for SortableList): the controlled-rollback sweep's generic recipe broke observer-only usage (binds *Changed without echoing the parameter back). Added an 'unchanged from before this interaction' no-op guard alongside the echo check — only a value differing from BOTH is a genuine, distinguishable veto. PickList's set-based guard is additionally gated behind an actual in-flight push, so a pure external reorder with no prior interaction still adopts correctly.
- **TimeWheelPicker**: a genuine transition from a real Value to null (a real clear/veto) now also resets the seed fallback (_ownLastEmitted), distinct from the pre-existing #68 'Value has always been null' observer contract, which stays untouched.
- **Scrollspy**: an empty/null ActiveId can now actually clear _currentActiveId (was previously silently ignored); the programmatic scroll stays scoped to non-empty ids.
- **Window**: RestoreFocus now runs AFTER OpenChanged.InvokeAsync (and only when the close was accepted), so a controlled veto no longer moves focus away from a window that stays open.

New/extended tests for all 6; full suite 4965 green, Release `-warnaserror` clean.

Brain2k-0005:
@codex review

Pushed 0745cfce — all 3 round-17 findings addressed:

- **TimeWheelPicker**: the accepted-echo branch (Value == _ownLastEmitted) now also records _lastSyncedValue, so a LATER genuine clear-to-null after an accepted commit is correctly detected (previously _lastSyncedValue stayed null through the accept, masking the later clear as the #68 ambiguous case).
- **Scrollspy**: clearing ActiveId now explicitly resets _pendingProgrammaticScroll (not just skips setting it), and OnAfterRenderAsync's guard treats empty as well as null and always consumes the flag — no more stale pending scroll firing for an empty id.
- **docs/Lumeo.Docs/wwwroot/registry.json**: synced — this artifact is populated by an MSBuild Copy target in Lumeo.Docs.csproj, which isn't part of Lumeo.slnx, so it was silently drifting across several rounds; now manually kept in sync after each RegistryGen run.

New regression tests for both code fixes, verified non-tautological. Full suite 4967 green, Release `-warnaserror` clean.

Brain2k-0005:
@codex review

Pushed 8ba69ce3 — both round-18 findings addressed:

- **Registry gate**: now mirrors the MSBuild Copy target (cp src/Lumeo/registry/registry.json -> docs/Lumeo.Docs/wwwroot/registry.json) before the freshness diff, and that path is added to PATHS — future drift of the top-level docs registry now fails CI automatically instead of needing a human to catch it.
- **PopoverTrigger**: EffectiveTriggerId (mirrors DialogTrigger/SheetTrigger's established _id pattern) now backs both the rendered id and the Space-suppression JS registration, so a consumer-splatted id no longer leaves the suppression targeting a non-existent element.

Investigated whether the same pattern recurs in AccordionTrigger/CollapsibleTrigger/TabsTrigger — it does, but their trigger id flows through a shared roving-focus registry + MoveFocus targeting in the parent context, so a narrow fix there would create a worse internal inconsistency than today. Left those for a dedicated, broader fix.

New regression tests, verified non-tautological. Full suite 4970 green, Release `-warnaserror` clean.

Brain2k-0005:
@codex review round 20 — fixed 3 findings from round 19:
1. CLI eject re-run gap (Commands.cs): a second `lumeo eject` after `add <component> --vendor` was a total no-op once `cfg.Standalone` was already true, leaving a kept satellite PackageReference stuck forever. The already-standalone branch now finishes stripping any package whose component is now vendored, with an E2E regression test reproducing the full first-eject-keeps -> add --vendor -> second-eject-finishes sequence.
2. PropsTable.razor rendered [CascadingParameter]s (FormField, InheritedDensity, a parent Context/Shell, ...) in the consumer-facing props table, advertising invalid markup. Now filtered via ApiParameter.IsCascading, with regression tests.
3. CI registry gate covered wwwroot/registry.json but not its sibling docs/Lumeo.Docs/wwwroot/registry-search.json (same MSBuild-target-not-in-Lumeo.slnx gap, round 18 vs round 19). Gate now regenerates and diff-checks both.

Full solution build (-warnaserror) clean, full CI-mirror suite (4970+44+30 tests across Lumeo.Tests/Lumeo.Cli.Tests/Lumeo.Docs.Tests) green. Please scan the whole codebase again for anything else.

Brain2k-0005:
@codex review round 21 — fixed 4 findings from round 20:
1. Cascader.razor: registered RegisterPreventDefaultKeys for the roving-focus nav keys (ArrowUp/Down/Left/Right, Home, End) while open, mirroring TreeSelect.OpenDropdown — same class of scroll-suppression gap.
2. ToggleGroup.razor: OnParametersSet used ValueChanged/SelectedValuesChanged.HasDelegate alone as the controlled signal, clobbering a callback-only observer's local toggle on every re-render. Now also compares against the pre-interaction parameter baseline (_lastValue/_lastSelectedValues); only a value distinct from BOTH the push AND that baseline still rolls back. Rewrote ToggleGroupControlledRollbackTests — the two 'veto re-renders the same starting baseline' cases are now indistinguishable from (and superseded by) the observer-only contract, plus two new tests prove a genuinely distinct override still wins.
3. RegistryGen: ConfirmButton drives OverlayProvider imperatively via IOverlayService and never rendered the tag, so the dependency scan never pulled in `overlay` — standalone/eject left ConfirmButton-only projects with no OverlayProvider once the package was stripped. Added an IOverlayService -> overlay dependency rule mirroring the existing FormFieldContext -> form rule.
4. SelectContent.razor: the search input's keydown bubbled to the listbox's own identical handler (no stopPropagation), double-firing OnSelect per keystroke — a no-op net toggle in Multiple mode, a double ValueChanged in Single mode. Added @onkeydown:stopPropagation, mirroring Cascader/TreeSelect's own search inputs.

Full solution build (-warnaserror) clean, full CI-mirror suite green: Lumeo.Tests 4975, Lumeo.Cli.Tests 45, Lumeo.Docs.Tests 30, Lumeo.RegistryGen.Tests 33, Lumeo.SourceGenerators.Tests 32. Please scan the whole codebase again for anything else.

Brain2k-0005:
@codex review round 22 — fixed 7 findings from round 21, plus one CI-blocking bug found while investigating a failed push:
1. ToggleGroup: round 20's observer-only fix over-generalized — a NON-null unchanged baseline is now always treated as a genuine, authoritative veto (rolls back); only a NULL unchanged baseline (the strongest signal Value was never bound at all) is treated as observer-only. Tests rewritten accordingly.
2. Input.razor: same null-only observer carve-out applied (no pre-existing conflicting test).
3. SignaturePad.razor: same finding, but deliberately NOT applied — SignaturePadValuePersistenceTests.Controlled_Parent_Rejection_Repaints_Canvas_To_Bound_Value already covers the identical null-baseline shape and asserts the OPPOSITE (erase, not preserve). The two scenarios are 100% indistinguishable from OnParametersSet's perspective — no signal exists to reconcile both. Comment-only change documenting why. Flagging this explicitly in case you want to weigh in on which behavior should win, or spot a signal we're missing.
4. Cascader.razor: registered Enter/Space in RegisterPreventDefaultKeys so a focused option button's native click-synthesis on Enter/Space no longer double-invokes SelectOption alongside the bubbled keydown handler.
5. Stepper.razor: added RegisterPreventDefaultKeys for Home/End/arrows against the tablist id (Stepper now implements IAsyncDisposable), mirroring every other roving-tabindex widget.
6. components.js positionFixed: RTL-aware start/end resolution via getComputedStyle(reference).direction, scoped to the horizontal-axis (side=top/bottom) usage only.
7. components.js positionFixed: containing-block compensation now runs BEFORE the flip/clamp overflow checks (so they measure the true on-screen position) AND after (to fold back any fresh values the clamp logic wrote), instead of only once at the end.
8. (self-discovered, not a Codex finding) scripts/algolia/generate-local-index.mjs: the docs-page existence check was case-sensitive-filesystem-dependent (existsSync against an assumed exact filename) — passed silently on Windows, failed on Linux CI for CTASectionPage.razor, dropping a component from the search index and failing the registry-freshness gate. Fixed to a case-insensitive filename lookup, mirroring RegistryGen's own PageFileExists.

Full solution build (-warnaserror) clean, full CI-mirror suite green: Lumeo.Tests 4981, Lumeo.Cli.Tests 45, Lumeo.Docs.Tests 30, Lumeo.RegistryGen.Tests 33, Lumeo.SourceGenerators.Tests 32. Please scan the whole codebase again for anything else.

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review round 23 — fixed both findings from round 22 (tools/lumeo-mcp/src/validate.ts):
1. Quote-aware tag scanning: the attribute-blob regex now alternates whole double-/single-quoted runs vs. a single plain non-quote/non-bracket char, so a `>` inside a quoted lambda (e.g. `Field="@(u => u.Name)"`) no longer truncates the tag match or leaves a self-closing element wrongly open on the ancestor stack.
2. Added a KNOWN_TYPE_PARAMS allowlist (T, TItem, TModel — the full set actually used via `@typeparam` across the library) so generic-component type-binding attributes are never flagged as unknown parameters.

Two new regression tests for the quote-aware scan, one for the allowlist; all verified non-tautological. `npm test` (tools/lumeo-mcp) green: 21/21. No .NET/registry changes this round. Please scan the whole codebase again for anything else.

Brain2k-0005:
@codex review round 24 — fixed all 5 findings from round 23:
1+2. tools/Lumeo.Cli/Commands.cs: `lumeo add --vendor` now aborts (instead of falling through to the OK summary) on a satellite-asset vendoring failure OR a failed vendored-dependency NuGet install — both previously recorded the component as installed anyway.
3. DatePicker.razor: OnParametersSet now hydrates Time from DateTimeValue.TimeOfDay (ShowTimePicker only), so binding a non-midnight DateTimeValue no longer loses the time on the next date pick.
4. RichTextEditor.razor: applied the same null-only observer-preservation carve-out as Input/ToggleGroup — verified against RichTextEditor's own existing controlled-rollback tests first (no conflict, unlike SignaturePad last round).
5. tools/lumeo-mcp/src/validate.ts: scoped the @typeparam exemption (round 22's fix) per-component via GENERIC_TYPE_PARAMS_BY_COMPONENT, so it no longer silently accepts typos like `<Button TItem="Order">` on non-generic components.

Ten new/updated regression tests, all verified non-tautological. Full solution build (-warnaserror) clean, full CI-mirror suite green: Lumeo.Tests 4983, Lumeo.Cli.Tests 47, Lumeo.Docs.Tests 30, Lumeo.RegistryGen.Tests 33, Lumeo.SourceGenerators.Tests 32; lumeo-mcp npm test 22/22. Please scan the whole codebase again for anything else.

Brain2k-0005:
@codex review

PR #351, opened 2026-07-08
fix: 4.1.1 — TreeView parent selection (#350) + GDPR consent hardening

### LINKED ISSUE #350: [Bug] TreeView: Issue with parent node selection
### Affected component or API

TreeView

### Reproduction

```markdown
Describe the bug
There is an error with the parent node selection logic in the TreeView component. The behavior when selecting a parent item is incorrect.
Steps to reproduce
1.Open the TreeView list.
2.Click on a parent node (e.g., "Documents" or "Photos").
3.Observe the selection state.
```

### Expected behavior

The parent node should be selected correctly according to the standard logic.

### Actual behavior

The parent node selection behaves incorrectly    [https://lumeo.nativ.sh/components/tree-view/](url)

### Blazor hosting model

WebAssembly (standalone)

### .NET version

10

### Lumeo version

4.1.0

### Browsers

Chrome / Chromium, Edge

### Browser console / build logs

```shell

```

### Anything else?

_No response_

Brain2k-0005:
### TreeView — fixes #350
A single click on a parent node co-selected **every** container node (selection was keyed by `Value`; value-less containers all share `null`). Selection is now tracked by node identity — public `SelectedValues` contract unchanged, values re-anchor across async `Items` refreshes, duplicate values no longer co-select, select vs expand independent (keyboard included), RTL-correct indentation. Regression tests with temp-revert proof; before/after verified on the docs demo.

### Consent — GDPR/TTDSG audit follow-up (code + live network audit)
- **Pre-consent third-party leak fixed (docs):** ECharts now self-hosted (pinned, immutable-cached) — fresh-context verification shows **zero** third-party requests before consent; the library's CDN-first default is unchanged and its self-host paths are documented on the Chart page + privacy policy.
- **Reopen surface:** new site-wide docs footer with a Cookie-settings control (works after dismissal) + Privacy link; pattern recommended on the ConsentBanner docs page.
- **Proof of consent:** stored record is now `{categories, timestamp, version}` with seamless legacy migration; new `PolicyVersion` parameter re-prompts on mismatch (choices prefilled, fail-closed while undecided); `DecidedAtUtc`/`DecisionPolicyVersion` exposed. Layer-1 Reject/Accept now have equal visual weight by default.

### Gates
Build `-warnaserror` 0/0 · suites **5,624 green** (Lumeo.Tests 5,477 · Docs 32 · RegistryGen 36 · SourceGen 32 · CLI 47) · registry + search regenerated · zero third-party pre-consent requests verified live-style.

## Summary by CodeRabbit

* **New Features**
  * Added a site-wide footer with “Cookie settings” to reopen consent preferences.
  * Consent now supports policy versioning so decisions can correctly re-prompt when policies change.
  * TreeView adds an “expand on row click” option (with chevron-only expansion when disabled).
* **Bug Fixes**
  * Hardened consent persistence and fail-closed behavior for malformed legacy/corrupt stored data.
  * Improved TreeView selection reliability across refreshes, duplicates, and lazy loading.
* **Documentation**
  * Updated privacy and component docs to reflect first-party/self-hosted chart assets and consent behavior.
* **Style**
  * Balanced the first-layer consent “Reject” and “Accept” button visuals.


Brain2k-0005:
@codex review

Brain2k-0005:
Round-1 findings all fixed (consent version re-evaluation + FOUC re-prompt sync, fully vendored + version-keyed chart assets with a clean fresh-context sweep, path-based reanchor, lazy selection resolution, scoped privacy wording) — @codex review

Brain2k-0005:
Added on top: TreeView row-click expand (VS Code pattern — row click selects + toggles parents by default, ExpandOnRowClick=false restores the chevron-only split; modifier clicks never flap folders; lazy parents load on row click). Browser-verified + 7 regressions. — @codex review

Brain2k-0005:
Round-2 findings all fixed — the four TreeView selection findings are consolidated into one documented selection lifecycle (seed/pending/resolved/superseded) instead of spot patches; PolicyVersion propagates on every parameter set; malformed legacy consent records fail closed; the chart plugin override keys are in the generated CDN manifest (14 deps). — @codex review

Brain2k-0005:
Round-3: TreeView pending carries are identity-preserving (path + value, ambiguity drops) and versionless proof records fail closed with a rejection latch — both tested. The ChartPage command finding is refuted: the bootstrap is written by default ('--no-bootstrap' is the real flag; '--write-bootstrap' does not exist in the CLI, System.CommandLine would reject it), so the documented command is correct as written. — @codex review

Brain2k-0005:
Round-4 all resolved: ambiguous carries drop immediately (kept only for unloaded lazy branches), row-click toggles before the awaited callbacks (stale-instance mutation in controlled trees fixed), the API scanner no longer emits private nested types (11 leaks removed from components-api.json), malformed category maps fail closed, and the self-host snippet carries the plugin keys. The AlertDialog CI failure was deflake-triaged 30/30 green locally (noise). — @codex review

Brain2k-0005:
Round-5 all fixed: expansion state flips synchronously while the lazy-load tail is drained in isolation (selection can no longer be delayed or broken by a loader — chevron/keyboard still surface loader errors), corrupt consent records fail closed AND clear the FOUC guard, and the docs no longer present EChartsSource as covering plugin extensions. — @codex review

Brain2k-0005:
Round-6 all fixed: timestampless proof records fail closed (legacy migration unaffected), path matches drop when the value is not unique among siblings (deliberate tightening of the round-3 survival semantics — ambiguity drops win; the four survival tests were flipped), and the per-chart EChartsSource is forwarded to the plugin charts' core load. — @codex review

Brain2k-0005:
@codex review

Brain2k-0005:
Round-7 all fixed: failed lazy expansions roll back and retry (selection intact), value seeds stay pending across unloaded lazy branches so the value contract binds every duplicate additively, and the privacy page now accurately enumerates which demo pages load third-party assets pre-consent (charts+fonts are self-hosted; Map/PDF/Scheduler/Editor/search are disclosed). — @codex review

Brain2k-0005:
Also in: tri-state parent checkboxes derive from seeded/lazy-loaded child state from the first frame (aria-checked=mixed on load, browser-verified), and the chevron is visually integrated into the row hover (transparent ghost, focus ring kept, still toggles without selecting). — @codex review

Brain2k-0005:
Round-9 both fixed: the lazy-load success path re-resolves the fresh node exactly like the failure rollback (shared resolver), and selections now track their origin — value seeds keep the every-match value contract across reloads while interactive selections keep identity carries with ambiguity drops; click converts seeded to interactive. Lifecycle doc updated, 4 regressions. — @codex review

Brain2k-0005:
Round-10 all four fixed: lazy loads start after the selection callbacks (cached completed-task loaders no longer bypass the ordering), the failure rollback clears IsLoading on the reanchored node, expansion targets get a value-verified positional fallback for duplicate-valued sibling parents (selection keeps strict drops), and the clicked node is re-resolved by path when OnItemClick reloads Items. — @codex review

Brain2k-0005:
Round-11 escalated instead of spot-fixed: the row-click/lazy/controlled-rebuild machinery is consolidated into ONE canonical post-await resolution model (single ResolveNode with ReferenceEquals -> value-verified path -> unique value -> DROP; no positional guessing in any mode — the round-10 fallback is deleted in favor of discarding re-fetchable loads over mis-attachment; deferral + change-detection contracts documented in the component). All four round-11 findings are solved by the model, with a 5-test ordering matrix. — @codex review

Brain2k-0005:
Round-12 solved as two new model pillars rather than spot fixes: (5) in-flight flag ownership — loads register path+value+sibling-snapshot entries and an orphan sweep clears rebuild-copied spinners whose slot has no live operation; (6) reanchoring reads from settled-state path snapshots, never the live list (covers in-place mutations). Non-lazy expands re-resolve like lazy ones; filtered displays rebuild on in-place swaps. 5 baseline-verified regressions. — @codex review

Brain2k-0005:
Round-13 fixed + we swept the whole component for model conformance ourselves: every path segment is now value-verified AND sibling-unique (ancestor swaps drop), in-flight loads are supersedable by any collapse (row/ArrowLeft/CollapseAll — the latter two found by our sweep, with regressions), filtered displays rebuild without requiring selection state, and ExpandAll completions have full per-row parity (re-resolve, supersession, orphan sweep). Conformance table in the commit. — @codex review

Brain2k-0005:
Round-14 answered structurally: UI state now lives in tree-owned NodeState entries keyed by once-per-rebuild identity resolution — async operations hold a stable handle no rebuild can invalidate. The in-flight registry, orphan sweep, supersession flags and every per-operation re-resolver (~540 lines, four subsystems) are DELETED; the consumer record flags remain the API as seeds + mirrors. All four round-14 findings dissolve without patches; rebuild-torture test included (controlled rebuild at every await point of a lazy row click). — @codex review

Brain2k-0005:
Round-15 solved as one contract: the seed comparison basis is now explicit (reference-matched -> store; fresh instances -> last incoming SEED, never our own mirror write-back), with a documented per-field precedence table and LoadToken gating so stale lazy completions can never clobber consumer-authoritative children (supersession bumps the token; success and catch both discard). Consumer resets collapse to pristine unloaded. 5 precedence tests + the self-trigger guard. — @codex review

Brain2k-0005:
Round-16 all four fixed within the precedence contract: empty authoritative lists replace (only null+unloaded resets), superseded load failures discard silently like their successes, the children seed advances on every adoption, and ExpandAll descends into consumer-authoritative children on token mismatch instead of skipping. Table updated, 4 revert-verified repros. — @codex review

Brain2k-0005:
Round-17 all four fixed: ExpandAll awaits already-in-flight loads (settle signal, supersession-tolerant) and swallows superseded failures per the per-node contract; the loaded-ness basis matches the documented rule everywhere (non-null list = authoritative incl. empty; plain-leaf guard test added); dropped deferred tails never start their loader. — @codex review

Brain2k-0005:
Round-18 all three fixed: the settle signal now marks the loader's TRUE start (reentrant ExpandAll from click callbacks cannot deadlock on a deferred tail; loading-without-signal branches are loaded by expand-all itself), supersession points complete the signal so parked waiters release immediately, and newly registered loader-returned empty branches mirror loaded-ness (no phantom chevron; leaf guard green). — @codex review

Brain2k-0005:
Round-19 fixed: the load attach is gated only on consumer supersession (LoadToken) — a newer expand-all generation no longer makes the older pass discard its loaded children before attaching; the superseded pass attaches once and stops descending while the newer pass takes over. Loader runs once, revert-verified concurrent test. — @codex review

Brain2k-0005:
Round-20 fixed (superseded expand-all attach now recomputes tri-state like the live path) and the overlay-exit test family is hardened against CI scheduler starvation (transient exit-class asserts are synchronous on the close render; unmount polls at the 10s module ceiling; 23 tests across 11 files, family verified 3x green — this addresses the two CI-only flaps without weakening coverage). — @codex review

Brain2k-0005:
Round-21 all three fixed: fresh lazy-child state is path-seeded at attach (identity-matchable across immediate rebuilds), value-equal SelectedValuesChanged echoes no longer re-seed (multiset comparison; different values still do), and versionless-rejected consent records expose no phantom decision timestamp. — @codex review

Brain2k-0005:
@codex review

PR #355, opened 2026-07-11
feat(ci): trust gates — docs parameter parity + public API baselines

Brain2k-0005:
Phase 1 of the trust campaign: make documentation drift and accidental API breaks structurally impossible.

## Docs/source parameter parity gate
`scripts/docs-parity/check.py` (stdlib-only) diffs every registry component's `[Parameter]` surface against the parameter rows in its docs API tables and fails CI on any gap. Allowlist entries require an honest reason and go stale loudly. Bring-up fixed 13 pages with genuinely missing rows; the 44 allowlisted entries are DataGrid-internal renderer sub-components never used outside the implementation.

## Public API baselines
`Microsoft.CodeAnalysis.PublicApiAnalyzers` (5.6.0) on all 11 non-icon shipped packages, `PublicAPI.Shipped.txt` baselines capturing the current 4.2.0 surface, enforced by the existing `-warnaserror` build. Gate proven live (fake public member → RS0016 → removed). Icon packs excluded: machine-generated, append-only surface (~51k members). Deprecation policy documented in CONTRIBUTING.

Gates: full build `-warnaserror` 0 errors with analyzers active, parity check exit 0, suites 5,885 green.

@codex review

## Summary by CodeRabbit

* **Documentation**
  * Expanded and refined component API reference tables (e.g., Combobox, DataGrid, DatePicker/DateRange/DateWheel, Input, Menubar, NumberInput, PasswordInput, PromptInput, Segmented, Sheet, Table, TimePicker, Textarea, plus several others).
  * Added/clarified new properties, events, and variant-specific guidance (e.g., focus behavior, disabled states, visibility controls, and child-item options).

* **Quality Improvements**
  * Added automated documentation parity validation to keep documented props/parameters in sync.
  * Introduced public API compatibility tracking for source projects and updated public API baselines.
  * Ensured Word import style map uses platform-independent line endings and added a dedicated test.


Brain2k-0005:
Round 1 addressed (see commit for detail):

- **QRCode casing (P1 + CodeRabbit critical)** — case-insensitive page index; casing drift warns instead of failing; the repo-wide sweep against `git ls-files` found QRCode as the only mismatch. CI is green again.
- **Stale-row reverse check** — documented rows without a source parameter now fail as `STALE_DOC_ROW`; bring-up caught two real drifts (Rating/SegmentedOption `Icon` → `IconContent`), fixed in the docs.
- **Chart subpages derived from source** — a missing/mis-cased chart page is now a `MISSING_DOCS_PAGE`, not silently tolerated.
- **Allowlist scoped by declaring type** — `slug/declaringType/param` with validation that the type really declares the parameter; the 44 slug-scoped entries became 59 precisely-scoped ones (all DataGrid-internal renderer sub-components).

Each new failure mode was proven live (temp-break → observe → restore). Gates: build -warnaserror 0 errors, parity exit 0, suites green.

@codex review

Brain2k-0005:
Round 2 addressed: the documented-rows set is now scoped by declaring type (8 more mis-filed pages surfaced and fixed), and the Linux-only RS0016/RS0017 failure is gone — the multiline public const made the baseline line-ending-dependent; it is now built from 
-joined literals with a pinning test and the baseline normalized.

Gates: build -warnaserror 0 errors, parity exit 0, suites green.

@codex review

Brain2k-0005:
Round 3 addressed: .razor.cs code-behind types normalize as a unit, dotted Component.Prop prefixes now win as the row owner, and rows owned by types outside the component file set no longer satisfy any check (one masked gap surfaced and fixed: SpeedDial.Disabled). Gates: build -warnaserror 0, parity exit 0, suites green.

@codex review

Brain2k-0005:
Round 4 addressed: the stale-row direction honors dotted owners like the forward direction (one more real drift surfaced and fixed: Column.Groupable named a nonexistent type), TimePicker Min/Max rows are scoped to the List variant, and the duplicate Rating IconContent row is gone. Gates: build -warnaserror 0, parity exit 0, suites green.

@codex review

PR #356, opened 2026-07-11
feat(a11y): keyboard-test families for all interactive components + 5 product fixes

Brain2k-0005:
Phase 2 of the trust campaign: close the keyboard-coverage gap the accessibility matrix made visible.

## What
- Keyboard-test families for **all 42 interactive components** that lacked them (registry `testCoverage.keyboard` 61/164 → **103/164**; the remaining 61 are passive layout/display primitives where keyboard tests would assert nothing).
- **5 product fixes** the audit surfaced: PivotGrid drill-down cells (focusable + Enter/Space, conditional on a wired handler), DatePicker Enter-commit bubbling into the popover trigger (reopened the calendar), ScrollArea keyboard reachability (APG scrollable-region, new optional `AriaLabel`), Dock roving tabindex (APG toolbar, reusing Toolbar's interop), and PullToRefresh honestly documented as a gesture-only gap rather than an invented fix.
- RegistryGen now counts a dedicated `*KeyboardTests` file as a keyboard audit — native-element components verify affordances (real `<button>`, no `tabindex` override) instead of dispatching synthetic KeyDown, which the content regex alone missed.

## Verification
~149 new tests, all suites green (5,810), temp-revert proofs on 12 tests across the four waves, full build `-warnaserror` 0 errors.

@codex review

## Summary by CodeRabbit

* **New Features**
  * Improved keyboard navigation, activation, and focus management across multiple components (including Dock, PivotGrid, DatePicker, Toolbar, and others).
  * **ScrollArea** now supports an optional accessible name via `AriaLabel`.
  * **PopoverTrigger** gained `SuppressActivationKeys` to prevent Enter/Space activation from toggling when child content already handles keys.
* **Documentation**
  * Refreshed keyboard/accessibility guidance and coverage details across the component catalog, including PivotGrid’s `--color-accent` CSS variable.
* **Tests**
  * Added and expanded keyboard-focused test suites to better validate tab order and native activation behavior.


Brain2k-0005:
Round 1 addressed — all 14 findings confirmed and fixed; the two registry data-integrity findings turned out to be library-wide scanner bugs (test misattribution touched 114/164 components and every .razor test was being dropped) — both scanners rewritten and the full diff audited.

Gates: build -warnaserror 0 errors, full suites green (5,813), registry regenerated.

@codex review

Brain2k-0005:
Round 2 addressed — all 11 findings fixed: activation keys are suppressed at the trigger (new opt-in PopoverTrigger.SuppressActivationKeys, since stopPropagation cannot be key-selective), PivotGrid registration survives consumer ids and late binding, and the registry scanners are honest again (folder-owned tests, [keyboard-gap] negative-audit marker, one-hop key-variable tracking, unified render-signal regex ending the stats/tests[] desyncs). Gates: build -warnaserror 0, regen, suites 5,860 green.

@codex review

Brain2k-0005:
Round 3 addressed: the suppressed trigger wrapper leaves the tab order entirely (no dead-end stop; origin discrimination is untestable in Blazor since KeyboardEventArgs has no target), toolbar nav keys exempt editable children at both the preventDefault and roving layers, both registrations reconcile symmetrically on parameter/id changes, and the keyboard-summary scanner now shares the emitter regex (dock summaries populated). Gates: build -warnaserror 0, regen, suites 5,868 green.

@codex review

Brain2k-0005:
Round 4 addressed — the round-3 patterns at their remaining sites: registered-id tracking on PopoverTrigger, editable-children exemption on Dock, splatted labels count as named ScrollArea regions, and switch-expression key matching is gated to keyboard-handler contexts. Gates: build -warnaserror 0, regen, suites 5,868 green.

@codex review

Brain2k-0005:
Round 5 addressed: the stale-listener leak was in JS (unregister re-looked-up an id the DOM no longer carried — handlers now store the element reference), editable toolbar children get a caret-boundary escape contract instead of a dead-end, and Dock targets its rendered id with runtime rebinding. Gates: build -warnaserror 0, regen, suites 5,876 green.

@codex review

Brain2k-0005:
Round 6 addressed: the button-trigger DatePicker had cancelling double toggles (keyboard could never open the calendar) — suppression is unconditional with both branches documented; Toolbar targets its rendered id with tracked rebinding; opaque editables stay arrow-opaque; a disabled ex-active item no longer holds the roving slot. Gates: build -warnaserror 0, suite 5,846 green (short-circuited gate).

@codex review

Brain2k-0005:
Round 7 addressed: stale tabindex cleared for aria-disabled/disabled roving candidates at the choke point, the wheel trigger variant is suppressed like the calendar branches, the a11y scanner no longer reads comments (repo-wide fix with audited diff), and suppressed triggers forward ARIA state to the focusable child via the AsChild slot mechanism. Gates: build -warnaserror 0, suites 5,857 + 43 green.

@codex review

Brain2k-0005:
Round 8 addressed: comment stripping is string-aware and covers line/block comments, dynamic a11y attributes (Razor role expressions, dictionary splats) are extracted so nothing legitimate is lost, and the suppressed trigger slot ships ARIA state without live toggle callbacks (double-toggle proven and fixed). Gates: build -warnaserror 0, suites 5,859 + 52 green.

@codex review

Brain2k-0005:
CodeRabbit follow-up addressed: computed-property role references resolve one hop into expression-bodied members — Result and Toast regain their alert/status roles in the generated metadata (regen diff limited to Result/Toast/Separator, 4 new extractor tests). Gates: build -warnaserror 0, suites 5,859 + 55 green.

@codex review

Brain2k-0005:
Round 9 addressed: excluded roving items are marked and re-examined so re-enabled elements rejoin, the calendar icon trigger carries the popup ARIA state alongside the input, and toolbar key rules re-register only on orientation changes. Gates: build -warnaserror 0, suite 5,859 green.

@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
Follow-up addressed: the suppressed slot strips the id before forwarding — the wrapper keeps sole ownership, ARIA attributes pass through, single-id DOM proven by test. Gates: build -warnaserror 0, suite 5,860 green.

@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
Fleet round addressed — every finding triaged and fixed at the root (or refuted with evidence); see the commit for detail. Gates: build -warnaserror 0, suites green, short-circuited.

@codex review

Brain2k-0005:
@codex review

PR #357, opened 2026-07-11
feat(toast): sonner-style stacking with hover/focus expansion

Brain2k-0005:
Sonner-style toast stacking: when several toasts share a position group they collapse into a stack — newest in front, older ones peeking behind with offset and scale (max three visible) — and hovering or focusing the group expands it back into the familiar vertical list. Opt-out via `ToastProvider.StackToasts`.

Two real bugs fell out of building this against a live browser:
- **The entrance animation never let go.** `animate-toast-in` (fill: both) stayed parked on every toast forever; a completed-but-filled CSS animation outranks any cascaded declaration, so the stacking transforms could never apply — and `animate-toast-out` never became the computed animation either, meaning **dismissals have been snapping instead of animating**. The class now drops on the entrance's real `animationend` (name-filtered Web Animations API listener + timer net for reduced-motion/JS-dead), and `Leaving` strips it immediately when a toast is dismissed mid-entrance.
- **The toast docs page mounted a second ToastProvider**, so every demo toast rendered twice with colliding element ids; removed in favor of the layout's global provider.

Verified in headless Chromium against the live docs page: 22/22 checks, including measured `scale(0.94)`/`translateY(10px)` on the first peeking toast, `opacity: 0` at depth ≥ 3, expand/collapse round-trips, and `animationName` transitioning to `toast-out` during dismissal. bUnit: 10 stacking/lifecycle tests; full suite 5,671 green; build `-warnaserror` clean.

@codex review

## Summary by CodeRabbit

* **New Features**
  * Added Sonner-style toast stacking (enabled by default) with hover/focus expand/collapse, per-position-group stacking, and a visible collapsed index cap (3); newest toast gets index 0.
  * Added `StackToasts="false"` to restore the classic non-stacked toast list.
  * Added a toast stacking demo and expanded toast API documentation.

* **Bug Fixes**
  * Improved toast entrance/exit lifecycle (including reduced-motion and RTL exit direction) and stabilized animation/class transitions during dismissals.

* **Tests**
  * Added toast stacking markup + interaction tests, admission-invariant regression coverage, and additional provider/cap/queue behavior tests.


Brain2k-0005:
Round 1 addressed, plus two user reports that arrived mid-review:

- **Peek direction** (user feedback): the stack now peeks AWAY from the anchored screen edge — bottom groups show older toasts above the front toast — with the absolute anchor following the peeking edge so mixed toast heights can't hide it.
- **Spam livelock** (user report "page crashes when spamming"): reproduced live — the MaxToasts eviction loop re-selected already-`Leaving` toasts; the synchronous idempotency guard meant two racing evictions spun a zero-yield loop that permanently froze the WASM main thread. Eviction now skips `Leaving` candidates and the enter-end interop tolerates disposed references. 100-toast bUnit spam test + 40-toast live spam check.
- **Depth≥3 exit flash** — `visibility: hidden` on the hidden slot (animations never override visibility, unlike opacity).
- **Mobile stack edge** — below `sm` everything bottom-anchors, so a media query forces the up-peek there.
- **RTL exit** — `toast-out-rtl` mirrors the dismissal; **reduced motion** — dismissals join the overlay family's 1ms-duration block, removal timing untouched.
- The unrelated `DataGridGroupCollapsePersistence` CI flap (second occurrence) is hardened per the family doctrine.

Verified in a real browser: 33/33 checks including measured peek-above offsets, spam interactivity, per-frame invisibility sampling during a hidden toast's exit, RTL direction and reduced-motion duration. Suites 5,705 green, build `-warnaserror` clean.

@codex review

Brain2k-0005:
Round 2 addressed — all ten findings plus the CI failure (same root as the MaxToasts finding): the bound is enforced via a capped pending queue with FIFO admission, the enter-callback interface is fully qualified and covered by a real CLI vendoring E2E test, hover and focus expand the stack independently, and leaving toasts keep a frozen index so hidden toasts can never be promoted into visibility mid-exit. The registry mis-attribution turned out to be LINQ .Select()/.Input() collisions in the scanner — fixed with regression tests. Browser-verified 44/44. Gates: build -warnaserror 0, regen, suites green, spam test 20x.

@codex review

Brain2k-0005:
Round 3 addressed — all five findings plus the CI diagnosis: the vendored Toast is free of internal Lumeo helpers (CLI E2E verifies), queued toasts accept in-place updates, MaxToasts clamps to >=1, focus transitions inside the stack get a sequence-guarded grace window, and overlapping exits always leave exactly one live front toast. The CI failure was a stale bUnit handle clicked across a re-render (atomic find+click now, 20x deflaked). Gates: build -warnaserror 0, regen, suites green.

@codex review

Brain2k-0005:
Round 4 addressed: the vendoring class is now dead by construction — the enter-end interop (which existed in no published package) is deleted in favor of the local fallback timer, and a new CLI E2E gate restores and compiles a vendored project against the live published package. Persistent toasts hold slots without starving the queue, exactly one toast owns the front slot through overlapping exits, and the scanner comment collisions are rephrased. Browser re-verify 44/44; suites green.

@codex review

Brain2k-0005:
Round 5 addressed: incoming persistents mount directly past the cap logic, MaxToasts is genuinely per position group now (accounting, eviction, queue and admission all scoped; tested with two groups), frozen indices are computed against the same live-only ranking the markup uses, and the deferred focus collapse is disposal-guarded. Gates: build -warnaserror 0, suites green (short-circuited).

@codex review

Brain2k-0005:
Round 6 addressed structurally: admission is now one canonical path — enqueue-first TryAdmit, ReconcileGroup as the sole mount/evict site owning all invariants (resolved duration decides persistence, slot accounting per group, re-reconciliation on position/persistence changes, FIFO admission), with mutation-into-reconciliation ordering instead of a mutex so bursts do not serialize behind 220ms exits. The entrance fallback timer is dispose-invalidated. A new invariant test asserts the per-group cap on every render of a deterministic burst. Gates: build -warnaserror 0, suites green.

@codex review

Brain2k-0005:
Round 7 addressed inside the admission model: destination-first reconciliation on group moves, live-only over-cap counting, and a synchronous Leaving stamp ahead of any await on the removal path. The burst invariant test now covers position moves and swipes. Gates: build -warnaserror 0, suites green.

@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
Round 8 addressed inside the model: reconciliation runs to a fixed point so bursts drain fully, queued retargets get destination-first ordering, and the change-triggering toast is protected from its own reconcile (deterministic survivor test). Gates: build -warnaserror 0, suites green.

@codex review

Brain2k-0005:
Round 9 addressed as the stage-3 move: admission state (position + persistence) freezes at entry against the then-current defaults — later default changes affect only future toasts, and only an explicit Update re-resolves (documented contract). Queued retargets respect the destination pending cap, over-cap evictions overlap, and the BCL-generic scanner collision is closed. Bonus: making the CLI test consume the real template imports surfaced the long-standing vendored-tag @using collision — the NamespaceRewriter now fully qualifies sibling tags, so rebranded installs compile. Gates: build -warnaserror 0, suites 5,685 + 130 + 41 green.

@codex review

Brain2k-0005:
Round 10 addressed: the resolved duration freezes at admission (queued toasts survive DefaultDuration changes; only Update re-resolves), and the vendoring qualification set now spans the full install set including transitive dependencies. Gates: build -warnaserror 0, suites 5,686 + 131 green.

@codex review

Brain2k-0005:
Fleet round addressed — every finding triaged and fixed at the root (or refuted with evidence); see the commit for detail. Gates: build -warnaserror 0, suites green, short-circuited.

@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
Fleet wave 3 addressed — findings triaged, fixed at root or refuted with evidence. Gates: build -warnaserror 0, suites green.

@codex review

Brain2k-0005:
Fleet wave 4 addressed — findings triaged, fixed at root or refuted with evidence. Gates: build -warnaserror 0, suites green.

@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
Wave 7 addressed — findings triaged and fixed at root. Highlights: the vendoring NamespaceRewriter no longer rewrites generic type arguments (P1, caught by the E2E compile test), toast stacking CSS is scoped to the viewport, and DataGrid grouped rows carry the correct aria-rowindex. Gates: build -warnaserror 0, suites green.

@codex review

Brain2k-0005:
Wave 9: the ToastViewport pause leak is closed at the single removal choke point (clearing the expanded-group marker the instant a position empties — DismissAll on a still-hovered default group no longer strands the next toast paused). Non-tautological test added. Gates: build -warnaserror 0, suites green.

@codex review

Brain2k-0005:
@codex review

PR #358, opened 2026-07-11
docs: registry-driven API tables + XML-doc backfill (drift extinction)

Brain2k-0005:
Phase 1b of the trust campaign: documentation that cannot drift.

- **XML-doc backfill**: 3,120 → 137 missing parameter descriptions (the rest live in components owned by open PRs, plus a handful of honestly-skipped unclear cases). Every summary verified against the component source; ships as IntelliSense and feeds the registry.
- **PropsTable sub-component rendering**: the registry's `api.subComponents` (58 composite components) now deserializes and renders; ButtonPage's path is byte-identical.
- **141 pages migrated** to registry-driven parameter tables — hand-written method/@ref, POCO and service tables stay. data-grid/tree-view/tag-input/agent-message-list deliberately stay hybrid (documented information-loss/name-collision risks).
- **Hybrid-aware parity gate**: a `<PropsTable>` usage marks that type's parameters documented while remaining hand tables keep the full forward+reverse check; bite proven by temp-removing a table (65 params re-flagged).

Gates: build `-warnaserror` 0 errors, parity exit 0 (141 skipped / 23 checked / 68 allowlisted), suites green.

@codex review


Brain2k-0005:
Round 1 addressed: the two chart summaries now state the real defaults (sweep found no further copy-paste cases), and the registry emitter textualizes see-langword/typeparamref/paramref inline tags instead of dropping them — the regen also repaired six components whose descriptions had been silently truncated. Excluded pages re-verified at zero PropsTable. Gates: build -warnaserror 0, parity exit 0, suites green (short-circuited).

@codex review

Brain2k-0005:
Round 2 addressed at the emitter/component layer so all 141 migrated pages benefit: single-point entity decoding (after tag-stripping — no double-decode path), derived implicit defaults for initializer-less value types, EventCallback rows deduped into the events section, and the named SelectItem descriptions backfilled verbatim from the removed hand rows. Gates: build -warnaserror 0, parity exit 0, suites green.

@codex review

Brain2k-0005:
Round 3 addressed: per-slug in-flight fetch dedupe (one HTTP request for sibling tables, tested), markdown-surface escaping of decoded XML with no double-decode chain, component-set-wide enum default resolution, and a structural Obsolete-message fallback so documented aliases survive the migration. Gates: build -warnaserror 0, parity exit 0, suites green.

@codex review

Brain2k-0005:
Round 4 addressed: synchronously-failing detail loads skip the in-flight cache entirely (retries verified by test), and code-span cells use pipe-only escaping so generic types render cleanly. Gates: build -warnaserror 0, parity exit 0, suites green.

@codex review

Brain2k-0005:
@codex review

PR #359, opened 2026-07-12
feat(ops): .NET 8 multitargeting, SourceLink/snupkg, standing eject gate

Brain2k-0005:
Phase 4 (wave 1) of the trust campaign: operational guarantees consumers can rely on.

- **.NET 8 + .NET 10 multitargeting** for every shipped package, with a real net8.0 test lane in CI (not just a compile check). The only degradation on net8 is IME composition detection in Command/PromptInput (`KeyboardEventArgs.IsComposing` is a net9 API) — documented in source and README.
- **Supply-chain metadata centralized**: SourceLink, deterministic builds, `ContinuousIntegrationBuild`, and snupkg symbol packages configured once in `Directory.Build.props/.targets` instead of 27 per-project copies. Consumers can step into Lumeo sources while debugging.
- **The NuGet-free eject promise is now a standing gate**: `eject-gate.yml` vendors *all* registry components into one fresh project and builds it — weekly, on every release, and on demand — failing loudly with the component list; a 5-component smoke variant runs on every PR. Locally proven: 164 components, zero Lumeo PackageReferences, green build in ~12 seconds.

Gates: full build `-warnaserror` on both TFMs (0 errors, 39 projects), representative packs verified (`lib/net8.0` + `lib/net10.0` + snupkg + split dependency groups), suites green, docs parity exit 0, actionlint clean.

@codex review

## Summary by CodeRabbit

* **New Features**
  * Added .NET 8 support across Lumeo libraries and components while retaining .NET 10 support.
  * Added automated “standalone eject” E2E verification to ensure no `Lumeo`/`Lumeo.*` package references remain after ejecting.
  * Improved published packages with enhanced SourceLink metadata and generated symbol packages.
* **Bug Fixes**
  * Improved keyboard navigation and prompt input behavior across supported .NET versions, including IME composition handling.
* **Documentation**
  * Updated README and contributing docs to reflect .NET 8+ support, multi-targeting, and standalone guarantee details.

PR #364, opened 2026-07-12
feat(trim): genuinely trimmable core — 83% smaller single-component apps (#354)

### LINKED ISSUE #354: WASM trimming: core assembly is not IsTrimmable — every app ships the full core DLL
## Finding

While re-measuring the documented core bundle size against 4.2.0 (fresh reference-app WASM publish, replicating the documented methodology), it turned out the core assembly is not IL-trimmed at all:

- `Lumeo.csproj` does not set `<IsTrimmable>true</IsTrimmable>`, so Blazor WASM's default `copyused` trim mode copies the assembly through wholesale.
- Verified directly: the published `_framework/Lumeo.dll` is SHA-256-identical to `lib/net10.0/Lumeo.dll` from the .nupkg, and a minimal app referencing only `<Button />` produces the byte-identical `Lumeo.dll.br` (687,723 bytes) as an app instantiating all 144 core components.

Consequence: any app using at least one core component ships the full 672 KB (brotli) core assembly. Satellite packages are unaffected — unreferenced satellite assemblies are dropped entirely, which is why the per-package split works.

The docs (BundleFacts) have been amended to state this honestly. This issue tracks the actual engineering work.

## Task

Evaluate marking the core assembly trimmable so member-level trimming works:

- Add `<IsTrimmable>true</IsTrimmable>` (or `TrimMode=full` readiness) to `src/Lumeo/Lumeo.csproj`.
- Audit reflection/dynamic usage first: JSON serialization of theme/preset options, `DynamicComponent`/`Activator` usage, interop marshalling, localization string lookup — anything reflection-based needs `[DynamicallyAccessedMembers]`/`[DynamicDependency]` annotations or explicit roots.
- Add a trimming-enabled reference-app publish to CI that fails on trim warnings (`-warnaserror` on ILLink) and byte-tracks the trimmed size.
- Re-measure and update BundleFacts once real per-usage trimming works — that would make "apps using a subset ship less" true for the core package and would be a genuine size win for consumers.

Measured baseline for comparison (2026-07-11, 4.2.0): `Lumeo.dll.br` 687,723 bytes (671.6 KiB) regardless of component usage.

Brain2k-0005:
Closes #354. The core assembly is genuinely trimmable now: 11 trim-hostile sites audited (9 annotated with DynamicallyAccessedMembers/RequiresUnreferencedCode propagated to the public generic surface, 2 refactored to a source-generated JsonSerializerContext), IsTrimmable + trim analyzers enforced by the -warnaserror build. Measured with a load-verified Button-only reference app: Lumeo.dll.br drops from 687,723 to 116,338 bytes — 83% smaller for single-component usage, where it used to be byte-identical regardless of usage. BundleFacts now tells the true story with dated before/after numbers. Stacked on #359 (multitargeting).

Gates: build -warnaserror 0 on both TFMs with trim analyzers, suites green on both TFMs, trimmed publish smoke renders, parity exit 0.

@codex review


Brain2k-0005:
Fleet wave 2 addressed — findings triaged, fixed at root or refuted with evidence (the snupkg push claim was checked against the NuGet client source). Gates: build -warnaserror 0, suites green.

@codex review

Brain2k-0005:
Fleet wave 3 addressed — findings triaged, fixed at root or refuted with evidence. Gates: build -warnaserror 0, suites green.

@codex review

PR #366, opened 2026-07-12
feat(trim): genuinely trimmable core — 83% smaller single-component apps (#354)

### LINKED ISSUE #354: WASM trimming: core assembly is not IsTrimmable — every app ships the full core DLL

Brain2k-0005:
Closes #354. (Re-opened as a fresh PR — the original #364 was auto-closed when its base branch `feat/ops-guarantees` merged into master; this carries the same commits rebased onto master, with the net8 multitargeting now inherited from the merged #359.)

The core assembly is genuinely trimmable now: 11 trim-hostile sites audited (9 annotated with DynamicallyAccessedMembers/RequiresUnreferencedCode propagated to the public generic surface, 2 refactored to a source-generated JsonSerializerContext), IsTrimmable + trim analyzers enforced by the -warnaserror build on both TFMs. The eject path was fixed so vendored components that reference the new JsonContext still compile NuGet-free (Serialization added to the substrate dirs).

Measured with a load-verified Button-only reference app: `Lumeo.dll.br` drops from 687,723 to 116,338 bytes — **83% smaller for single-component usage**, where it used to be byte-identical regardless of usage. BundleFacts now tells the true story with dated before/after numbers.

Gates: build -warnaserror 0 on both TFMs with trim analyzers active, suites 5,870 green, trimmed publish smoke renders, parity exit 0, eject smoke green. The merge into master resolved a PublicAPI.Unshipped conflict (union of this branch's ModelTypeInfo entry and master's a11y params).

@codex review

## Summary by CodeRabbit

* **New Features**
  * Forms now support trim-safe model snapshots through optional generated JSON metadata.
  * Query Builder serialization better preserves numeric and date values across cultures and rejects invalid data safely.
  * Added a WebAssembly trimming demonstration sample.

* **Bug Fixes**
  * Form-level validation errors now display correctly.
  * Improved preservation of dynamically used icons, overlays, and interop behavior in trimmed applications.

* **Documentation**
  * Updated trimming guidance and measurement results for Lumeo 4.3.0.


Brain2k-0005:
Round addressed: Form re-captures its snapshot on Model change; QueryBuilder preserves custom query values through the new JsonContext serialization; and oversized numeric literals now fail parsing (JsonException) instead of silently clamping to +Infinity and corrupting the query — a real latent bug the corrected round-trip test uncovered (its hand-built JSON was missing the polymorphic \$type discriminator, masking the number path). Tests extended for all three. The registry-version-bump finding is deferred: the branch stays 4.2.0 and bumps to 4.3.0 at the release bundling this campaign (a per-PR bump would collide across seven open PRs). Gates: build -warnaserror 0 both TFMs, suites 5,873 green, parity exit 0.

@codex review

Brain2k-0005:
Wave 8 — the recurring QueryBuilder serialization findings are consolidated at the root: the value converter dispatched via an open-ended type case-list that grew one type per review round; it now uses a closed Type.GetTypeCode switch (all numeric primitives incl. unsigned in one arm), explicit Enum/Guid arms with Enum.IsDefined, NaN/Infinity rejection, and double-widening for bit-exact float round-trips. A 44-case fuzz matrix (numeric boundaries, enum, guid, unicode, nested/empty groups) locks the surface so a fourth round finds nothing. All 4 findings resolved within it. Gates: build -warnaserror 0 both TFMs, suites 5,919 green, parity exit 0.

@codex review

Brain2k-0005:
@codex review

PR #378, opened 2026-07-18
feat(input): custom vertical steppers for type=number, replacing native spinners

Brain2k-0005:
## Summary
- Generic `Input` now renders its own themed vertical ▲▼ steppers for `type="number"` instead of falling through to the browser's tiny native spin buttons — reuses NumberInput's exact visual idioms (colors, borders, focus ring) so the two controls read as one family.
- Detection is case-insensitive on both the attribute key and value (`Type="number"` from Razor markup lands as `AdditionalAttributes["Type"]` — capital T, verified empirically — while a hand-written lowercase `type="number"` is equally valid HTML).
- Native spinner arrows are always hidden on a number input, independent of the new `ShowStepButtons` opt-out (default `true`); with it `false` the field renders as a clean bare number input with no steppers and no native spinner.
- `min`/`max`/`step` are read straight off the existing HTML attributes (no new typed params) — the buttons clamp/disable at the same bounds native ArrowUp/ArrowDown already respects.
- Extracted NumberInput's clamp / floating-point cleanup / step-decimal-digit math into an internal `NumberStepper` helper so both controls share the exact same numeric core, behavior-preserving (NumberInput's existing suite is unchanged and green).
- Reuses the existing `NumberInput.Increase`/`NumberInput.Decrease` locale keys (all 14 locales) for the button aria-labels — no new strings.
- No new Tailwind utility classes were introduced (every class already exists elsewhere in the codebase), so no CSS bundle rebuild was needed.

## Test plan
- [x] New `NumberStepperTests` — pure clamp / FP-cleanup (0.1+0.2 stays 0.3) / step-decimal-digit unit coverage.
- [x] New `InputNumberStepperTests` — steppers render for both `Type="number"` and `type="number"` casing; do not render for text/search/file/email/no-type; increment/decrement fire `ValueChanged` with the invariant string; respect min/max/step attributes and disable at bounds; `ShowStepButtons=false` renders no buttons but still hides the native spinner; localized aria-labels + `tabindex=-1`.
- [x] NumberInput's existing suite (`NumberInputTests`, `NumberInputKeyboardTests`) unchanged and green — confirms the `NumberStepper` extraction is behavior-preserving.
- [x] Clean build: `dotnet build Lumeo.slnx -warnaserror -m:1` — 0 real diagnostics.
- [x] Docs parity: `python scripts/docs-parity/check.py` — exit 0.
- [x] Full suite (excluding E2E/EjectGateFull): 6056 tests green on both net8.0 and net10.0, plus CLI/Docs/RegistryGen/SourceGenerators suites all green — no flaky retries needed.
- [x] Docs page updated with a "Number Steppers" example demonstrating `min`/`max`/`step` and `ShowStepButtons`.

## Summary by CodeRabbit

* **New Features**
  * Added themed ▲▼ step buttons for `type="number"` inputs (native spinners are no longer shown).
  * Introduced `ShowStepButtons` to optionally hide custom steppers while keeping real number input behavior.
  * Added `min`, `max`, and `step` handling with value clamping and correctly disabled buttons at bounds.
  * Improved fractional/decimal stepping accuracy to prevent drift on repeated increments.

* **Documentation**
  * Updated the Input component page with a “Number Steppers” demo and accessibility notes.

* **Tests**
  * Expanded unit and UI coverage for stepping, bounds, disabled/readonly behavior, rendering, and edge cases.


Brain2k-0005:
@codex review

PR #379, opened 2026-07-21
feat(gantt): Blazor-first Gantt v3 - Phase 1 read-only rendering + v2/v3 parity harness

Brain2k-0005:
## Summary

Phase 1 of the Gantt v3 rewrite: the imperative 887-line SVG renderer (`gantt-v2.js`) is being replaced by a Razor-first component tree (strangler pattern). **v2 stays shipping and untouched** — the only v2-file change on this branch is the additive `GanttTask.ParentId` field. v3 lives in `src/Lumeo.Gantt/UI/GanttV3/` under the working name `Gantt3` and swaps in only when the parity suite gates it (Phase 4).

### What's in Phase 1 (read-only rendering)

- `GanttScale` — pure C# port of v2's `VIEW_MODES` layout math (byte-matched: column widths, snap units, date↔pixel mapping incl. JS `Math.round` tie-break semantics, two-row header segmentation, DST-safe)
- `Gantt3` root + `GanttNav` (Today / prev / next / view-mode switcher) + `GanttTimeline` (two-row grouped header, virtualized rows, DOM+CSS bars, progress fill, milestone diamonds, today marker, `BarTemplate` slot)
- `GanttTree` pane — flat `GroupBy` grouping and new `ParentId` hierarchy, collapse/expand, cycle-safe row model (invalid parent chains render instead of vanishing)
- `GanttArrowLayer` — dependency arrows, routing math ported 1:1 to C# (`GanttArrowRouting`) incl. v2's unconditional backward-arrow behavior
- No drag/JS interop yet (Phase 2); all colors via CSS variables

### Parity harness (the swap gate)

ServerHost pages `/e2e/gantt-v2`, `/e2e/gantt-v3`, `/e2e/gantt-v3-tree` share one fixture; Playwright specs assert v2 DOM, v3 DOM **and** independently derived ground-truth values (±1px). The harness already paid for itself — it caught three real v3 bugs invisible to bUnit:

1. `Color="row.Color"` bound the literal string instead of the expression (custom bar colors silently dead)
2. `aria-expanded` rendered as a bare boolean attribute instead of `"true"/"false"`
3. Dependency arrows rendered ~56px (header height) too high — SVG overlay spans the outer canvas, bars get the header offset via DOM nesting

Intentional deltas are pinned and documented in-spec (milestone label ~2px, nav is v3-only).

### Public API

All six new public components/types carry `[EditorBrowsable(Never)]` and live in `PublicAPI.Unshipped.txt` only (161 lines). **They must not be promoted to Shipped before the Phase-4 rename** (`Gantt3` → `Gantt`), otherwise the rename becomes an RS0017 break.

### Tests / gates

- Unit: 89 GanttV3 tests, full suite 6162/6162 green, `-warnaserror` clean on net8.0 + net10.0
- E2E: 29 functional parity/nav/tree specs (stable across repeat runs), 8 visual snapshots per mode/route
- CI: `e2e.yml` now boots the test ServerHost (readiness-polled) so the parity specs run on PRs

### Next phases

2. Drag interactions via thin JS layer (move/resize/progress, ghosts, snapping) behind the same parity gate
3. REUI-class additions: infinite scroll, summary bars, row reorder, hoistable `GanttState`, templates
4. Swap: rename to `Gantt`, keep v2 in-tree unreferenced for one minor release as rollback

## Summary by CodeRabbit

* **New Features**
  * Added Gantt v3 components (read-only navigation, zoom controls, today highlighting, task/milestone bars, progress, dependency arrows, timeline header).
  * Added optional hierarchical tree pane support driven by `ParentId`.
  * Added a Gantt v3 preview E2E documentation page.
* **Bug Fixes**
  * Preserve `ParentId` across edit callbacks and JS interactions.
  * Improved sticky-header and horizontal scroll-to-today behavior, including RTL.
* **Documentation**
  * Added missing Gantt navigation/expand/collapse labels.
* **Tests**
  * Expanded rendering, parity, and E2E coverage; improved failure log artifacts.


Brain2k-0005:
@codex review

All 8 findings from the previous round are addressed and pushed (initial scroll-to-today incl. deterministic E2E latch, ColumnWidth threading through all geometry, VisibleRange recompute on task-set change, collapsed-group leak, ParentId round-trip preservation on all callback surfaces, milestone diamond bounding box, tree a11y labels, sticky-header scroll container), plus the CodeRabbit round (BarTop reuse, Dependencies sequence equality, E2E wait contract, MD040).

Brain2k-0005:
@codex review

Round 2 findings are all addressed and pushed (6 commits, a8f81dd9..04a106f6): ParentId in the task hash gate + click payload, flat-group collapse leak + empty-GroupLabel truthiness, CurrentCulture month labels, sticky-header restructure (real fix this time, with a tall-fixture E2E spec proving it), group section rows rendered in the timeline, tree-pane virtualization, browser-local today, and deterministic visual snapshots.

Brain2k-0005:
@codex review

Round 3 findings all addressed and pushed: timeline Virtualize now attaches to the outer shared scroll pane (with hard-count E2E for both panes and a live-verified sticky-left tree pinning fix), empty-to-populated recenter, TodayHighlight decoupled from viewport centering, v2-parity empty state and milestone tooltip normalization, unit-aligned Month/Year Today-recenter, and RTL scroll normalization with real DirectionProvider E2E coverage.

Brain2k-0005:
@codex review

Round 4 findings all addressed and pushed: sticky tree header spacer, tree-width-aware scroll centering, virtualized dependency arrows (rAF-throttled scroll tracking), corrected RTL scroll-convention probe, keyboard-focusable bars with localized labels, all Gantt strings translated for every shipped locale with a completeness guard test, CustomClass forwarding, viewport-center-preserving view-mode switches, and physical date order plus logical-edge tree pinning under RTL.

Brain2k-0005:
@codex review

Round 5 findings all addressed and pushed: pointer clicks now reach OnTaskClick while preserving the tooltip's own tap-to-pin and Escape handling (with a live-browser E2E proving real event bubbling), button semantics only when a click delegate exists, corrected RTL centering (root cause: timeline root rendered wider than its content, corrupting RTL maxScroll), view-mode switches preserve the live scroll center, localized progress text across all shipped locales, interval-intersection arrow culling that keeps window-crossing dependencies, deduped vertical scroll reporting, and RTL pre-range scroll targets clamp to the earliest edge. Visual snapshots are now clock-independent on both routes via a frozen browser date.

Brain2k-0005:
@codex review

Round 6 findings all addressed and pushed: continuous month interpolation now clamps to each month's real day count (Feb/leap-year/31-day cases pinned), arrow-culling window refreshes on row-count and pane-resize changes, RTL convention labels corrected and exhaustively documented, Space activation no longer scrolls the viewport (registration reconciles on delegate changes), and the v2 click payload merges only ParentId onto the renderer-normalized task.

Brain2k-0005:
@codex review

Round 7 findings all addressed and pushed: the continuous month inverse is now bijective against each month's real day count, scroll-host tracking rebinds on host identity changes, the browser-local today refreshes on server-day advance (no cross-timezone render thrash) and on every Today action, and direction flips on a mounted chart re-center the viewport. The visual parity suite also runs locally again (solution-anchored snapshot path) with all eight baselines passing.

Brain2k-0005:
@codex review

Round 8 findings all addressed and pushed: tree-pane visibility changes and direction flips both re-center on the live viewport center (a manual pan survives either), header synchronization reconciles on scroll-host identity changes, bar geometry truncates task times to date-only matching the shipping renderer's parseDate, and non-milestone tasks with End before Start are filtered out with their dependencies skipped, as v2 does.

Brain2k-0005:
Thanks — round 9 acknowledged: all seven findings are triaged as real and are tracked as the FIRST work item (T0) of the immediate follow-up cycle on this component, which continues right after this PR merges (the strangler campaign's next phase branches off master and goes through the same review loop, so these will come back across your desk shortly). Given nine rounds with severity trending from P1 architecture issues to same-day-time-normalization edge cases, we're merging this phase now and closing the tail in the next one.

Brain2k-0005:
Update to the previous note: we are closing all seven round-9 findings in THIS PR after all (quality bar: the loop runs until a clean round), not deferring them. Fix wave incoming.

Brain2k-0005:
@codex review

Round 9 findings all addressed and pushed: duration validity now shares one date-normalize-then-filter step between the row model and the range computation (same-day inverted times are valid as in the shipping renderer, filtered tasks no longer skew the range), the header transform clears on sync unregister, async-populated ranges center on their tasks when today falls outside, tooltips show the clamped progress value, and theme-service-driven direction changes trigger the same recenter as provider-driven ones.

Brain2k-0005:
@codex review

Round 10 findings all addressed and pushed: virtualized rows now render an in-flow row box of the exact row height (visual baselines pass byte-unchanged, confirming zero geometry drift), parent-driven and toolbar view-mode changes share one capture-center-rebuild-recenter path, the default tree-pane visibility derives from the filtered task set, and Month/Year Today-recentering preserves the rendered column count across variable-length calendar units.

Brain2k-0005:
@codex review

Round 11 findings all addressed and pushed: all recenter triggers (view-mode, tree-pane, direction — toolbar-, parameter- and theme-service-driven) now flow through one capture-apply-recenter pipeline with a single live-center read under the old geometry, sub-day mode switches align range boundaries to their hour-step while keeping the user's exact panned position as the target, and emptiness transitions in either direction re-center the viewport.

Brain2k-0005:
@codex review

Round 12's finding is addressed and pushed: the tasksChanged x viewModeChanged matrix is now explicit — when both change in one render, the pipeline captures the live center under the old geometry, applies the mode, rebuilds the range from the new task set under the new mode, and recenters on the tasks for emptiness transitions or the preserved center otherwise. All four matrix cells carry dedicated tests.

Brain2k-0005:
@codex review

Round 13's finding is addressed and pushed: effective column-width changes are folded into the recenter trigger set — width-only and width-plus-mode changes both preserve the viewport center, with the live-center capture decoding under the fully-old geometry (old mode and old width). The reconcile matrix now covers tasks, mode and width, each cell tested.

Brain2k-0005:
@codex review

Round 14's findings are addressed structurally: the parameter-reconcile machinery is consolidated into a single snapshot-diff owner (GanttViewportReconciler) — one input snapshot per parameter set, one diff decision, live-center capture under the fully-old geometry BEFORE any state commit (no partially-reconciled intermediate frames on Blazor Server), one-shot expiring scroll intents instead of persistent targets, emptiness classified on the filtered task set, and the post-switch effective column width recorded after the mode change. Net deletion of the accreted per-parameter tracking; all prior matrix tests unchanged and green.

Brain2k-0005:
@codex review

Round 15 findings all addressed and pushed: reconciles carry a generation — a capture superseded during its await abandons its commit and the newest reconcile wins regardless of resolve order; scroll applies consume only the request id they captured before awaiting; the last raw-count consumer now uses the renderable set (swept, none remain); and virtualized bars are keyed by task identity.

Brain2k-0005:
@codex review

Round 16 findings all addressed and pushed: no-op theme notifications no longer claim a reconcile generation (direction compared against the last snapshot first), superseded reconciles suppress their callbacks and side effects via an explicit committed/superseded result, row keys are namespaced so task ids can never collide with synthetic group keys, disjoint task-range replacements recenter while combined tasks-plus-mode behavior stays untouched, the scroll-center capture takes the direction convention explicitly instead of trusting the live DOM under a theme flip, and bars unregister immediately when a key-handler registration resolves after disposal. Each fix was individually revert-verified against its own regression test.

Brain2k-0005:
@codex review

Round 17 findings all addressed and pushed: no-op parameter renders no longer claim a reconcile generation (snapshot diffed first, symmetric with the theme path), disposal claims a final generation so suspended reconciles can never commit afterward, the vertical-scroll-tracking registration and the key-handler path both recheck disposal and delegate presence after their awaits, and zero column-width/bar-height overrides fall back to defaults exactly like the shipping renderer's truthy check. Each fix revert-verified against its own test.

Brain2k-0005:
@codex review

Re-triggering — the previous request from last evening appears to have gone unanswered. Round 17's fixes are on the branch (no-op parameter renders don't claim generations, disposal invalidates suspended reconciles, late-interop rechecks on all registration paths, zero dimension overrides fall back like v2). A test-only deflake of two CI-flaky Toast tests is landing shortly as well.

Brain2k-0005:
@codex review

Requesting the round again — the two prior triggers appear to have gone unanswered. Since the last review round: round 17's reconcile lifecycle fixes (no-op parameter renders claim no generation, disposal invalidates suspended reconciles, late-interop rechecks, zero-dimension defaults) and two test-only deflakes of a CI-starvation Toast race (root-caused via a 1ms-timer reproduction, exit window pinned per the suite's established doctrine).

PR #380, opened 2026-07-22
fix(datagrid): sorting ignored while grouping is active

Brain2k-0005:
## Bug

With grouping active, clicking a column header to toggle Ascending/Descending
sort appeared to do nothing.

## Root cause

The grouping pipeline (both single-level and multi-level) built its group
list via `.GroupBy(...).OrderBy(g => g.Key, StringComparer.CurrentCulture)` —
the group ORDER (which group section/node appears first) was hardcoded
ascending, completely independent of the grid's active sort state.

Row order *within* a group was never actually broken: the sort is applied to
the item list before grouping runs, and LINQ's `GroupBy` preserves each
group's members in their original relative order. The bug is specifically
that sorting the **grouped column itself** had zero visible effect — every
row in a group shares that column's value by definition, so "sort within the
group" is necessarily a no-op for that field, and group order (the one place
the sort direction could show up) was hardcoded.

## Fix

Group order now follows the active sort entry for the grouping field:
Descending reverses the group run, and no active sort on that field falls
back to the original ascending-by-key default. This is the semantic used by
grouped columns in TanStack Table, AG Grid, and MUI DataGrid — sorting the
grouped field controls group order; sorting any other field controls order
within each group (already working, unaffected by this change).

Applied to both single-level and multi-level (per nesting level
independently) grouping. Server mode shares the same fixed code path via its
existing regroup-the-delivered-page routine, so it benefits without any
separate change.

## Tests

New regression suite covers:
- Sorting the grouped column Ascending/Descending reverses group order;
  clearing the sort restores the default order.
- Sorting a non-grouped column still reorders rows within each group in both
  directions (regression guard for the part that already worked).
- Plain ungrouped sorting is unaffected.
- Sort state survives toggling grouping on and off in both directions.
- Multi-level grouping: each level honors its own sort direction
  independently.
- Server mode.

## Gates

- Release build (`-warnaserror`) of the grid library: 0 warnings, 0 errors,
  both target frameworks.
- Focused grid test filter: all passing, no regressions.
- Full unit suite: all passing.
- Component registry regenerated (doc-comment/test-discovery metadata only).

## Summary by CodeRabbit

- **Bug Fixes**
  - Grouped DataGrid sections now follow the selected ascending or descending sort direction.
  - Multi-level grouping honors sorting independently at each grouping level.
  - Sorting continues to work when switching between grouped and ungrouped views, including Server Mode.
  - Sorting non-grouped columns still reorders rows within their groups without changing group order.

- **Tests**
  - Expanded automated coverage for grouped sorting scenarios.

### DIFF
--- a/docs/Lumeo.Docs/wwwroot/registry/cdn-deps.json
+++ b/docs/Lumeo.Docs/wwwroot/registry/cdn-deps.json
@@ -1,6 +1,6 @@
 {
   "version": "1.0",
-  "generated": "2026-07-18T08:12:42.4994039Z",
+  "generated": "2026-07-22T14:36:14.2184664Z",
   "deps": [
     {
       "key": "mapLibreJs",
--- a/docs/Lumeo.Docs/wwwroot/registry/data-grid.json
+++ b/docs/Lumeo.Docs/wwwroot/registry/data-grid.json
@@ -3065,8 +3065,8 @@
   },
   "testCoverage": {
     "tier": 4,
-    "files": 42,
-    "tests": 290,
+    "files": 43,
+    "tests": 299,
     "relatedFiles": 1,
     "render": true,
     "behavior": true,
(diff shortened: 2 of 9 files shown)


Brain2k-0005:
Standing down per owner clarification: sorting is applied within each stable group by design, not a bug. Closing without merging.

PR #381, opened 2026-07-22
fix: sidebar scroll/list-validity, NumberTicker updates, drawer form height

Brain2k-0005:
## Summary

Fixes for four documented upstream issues, plus one investigation that did
not reproduce.

### 1. NumberTicker Value updates (no fix needed)
Reported: NumberTicker allegedly ignores `Value` updates unless the consumer
forces an `@key` remount. Investigated at the library level and could not
reproduce it — the JS tick registration always cancels the prior animation
frame and re-registers unconditionally on every call, and the component
re-ticks on every genuine `Value` change to the same instance regardless of
remounting. Added regression tests guarding this behavior (asserting the JS
interop receives the correct from/to pair across repeated updates) since none
existed before.

### 2. SidebarContent scroll defaults
`SidebarContent`'s root defaulted to `overflow-auto` (both axes). Nav content
routinely grows taller than the viewport, so the vertical scrollbar appears
and steals width from the content box — every full-width entry then reports
as slightly too wide for the now-narrower box, triggering a spurious
horizontal scrollbar too. Now scrolls only the vertical axis
(`overflow-y-auto`/`overflow-x-hidden`, the same pair the sidebar's own root
already uses) and reserves the scrollbar's track space up front
(`scrollbar-gutter:stable`) so the box doesn't reflow width-wise when the
list crosses one page.

### 3. SidebarGroupLabel / SidebarSeparator inside SidebarMenu
Both render a `<div>` root, which is invalid directly inside `SidebarMenu`'s
`<ul>` — using either to break up a single menu's items into visual
sub-sections required hand-wrapping in `<li role="presentation">`.
`SidebarMenu` now cascades a marker both components read to auto-detect that
placement and wrap themselves accordingly, with no opt-in required. Used in
their canonical placement instead (`SidebarGroupLabel` as a sibling before
`SidebarMenu` inside `SidebarGroup`; `SidebarSeparator` between two
`SidebarGroup`s — every example in the docs already does this), both render
their plain `<div>` exactly as before. Added explicit composition guidance to
the docs page.

### 4. Drawer collapses tall content
A consumer's form inside `DrawerContent` became unreachable for tall content.
Root cause, confirmed with a real headless-browser render: the panel caps its
own height but never gave itself a scroll fallback — once content exceeds
that cap, the excess renders past the panel's bottom edge with no way to
scroll to it. The header stays visible, but everything below a point,
including the footer and its action buttons, renders completely off-screen
and is permanently unreachable. `overflow-y-auto` on the panel turns that
same overflow into an internally-scrollable region; the header/footer get
`shrink-0` defensively.

## Test plan
- [x] New regression tests per item (NumberTicker, SidebarContent, Sidebar
      list-validity, Drawer) — all passing
- [x] Full focused filter (Sidebar/Drawer/Motion + NumberTicker): 146/146
- [x] Full unit suite: 6083/6083
- [x] Release build (`-warnaserror`), both touched projects + full solution: 0/0
- [x] Component registry regenerated (final commit)

## Summary by CodeRabbit

* **Bug Fixes**
  * Drawers now keep tall content scrollable within the panel, with swipe/snap gestures no longer interfering with in-panel scrolling.
  * Sidebars now use vertical-only scrolling with stable layout, and render valid menu list markup for group labels and separators (avoiding nested list artifacts).
* **Documentation**
  * Clarified sidebar API guidance for supported nesting/cascading and placement/override behavior.
* **API Updates**
  * Updated sidebar-related cascading state and made `AsListItem` overrides nullable.
* **Tests**
  * Added/expanded coverage for drawer scrolling/drag-handle behavior, sidebar scrolling/HTML validity, and NumberTicker reticking.


Brain2k-0005:
@codex review

All four findings from the first round are addressed and pushed: swipe-to-dismiss only engages when the panel is at its closing scroll boundary at gesture start (content scrolling wins otherwise, handle gestures always dismiss), the scrollbar gutter applies only in the expanded sidebar state, the list-item wrap is constrained to direct SidebarMenu children, and scrolling drawers reserve padding so focus rings stay visible. The registry freshness gate is also fixed.

Brain2k-0005:
@codex review

Round 2 findings addressed and pushed: the content-vs-dismiss decision now latches for the entire touch (a gesture that starts away from the closing boundary stays content-owned until the next touch begins — no mid-scroll re-arm), and touches starting on the drag handle bypass the scroll gate entirely and always arm dismiss.

Brain2k-0005:
@codex review

Round 3 findings addressed and pushed: opening-direction snap drags honor the scroll boundary rule symmetrically, gesture ownership is latched before the activation threshold on every frame (closing a direction-reversal bypass), and the snap panel rests via offsets with the transform only transient during drags — at rest the computed transform is none, so fixed-position popovers inside the drawer position and render correctly again (verified in real Chromium).

Brain2k-0005:
@codex review

Round 4 findings all addressed and pushed, anchored on an explicit gesture state model (rest / dragging / transitioning / content-owned) documented at the top of the gesture code: content-owned touches short-circuit move, end and velocity paths alike; the ownership latch provably precedes every activation-threshold early return; the panel enters drag representation lazily on the first owned move (content-owned touches never leave rest representation); rest conversion swaps representation with the transition disabled and a forced reflow (no settle flash); and snap unregistration clears the resting offsets. Verified by dispatching real TouchEvents in headless Chromium — twenty assertions across four scenarios, looped three times.

Brain2k-0005:
@codex review

Round 5 findings addressed and pushed: a touch that ends without any movement is a complete no-op on the representation (documented as its own row in the state table), and pending settle transitions carry a generation that unregistration invalidates — with registration also defensively tearing down any stale live registration first. Verified with real touch events in headless Chromium across tap, mid-settle-unregister and mid-settle-reregister scenarios.

Brain2k-0005:
@codex review

Round 6 findings addressed and pushed: transitions are disabled lazily at real drag start (taps and content-owned touches never interrupt a running transition), the opening frame is guarded by the same generation token as settles, the drag handle is sticky at the panel top so it never scrolls away (boundary gating unchanged), and registration saves and restores any consumer-supplied inline offsets on teardown. Verified in real Chromium including regression re-runs of the prior rounds' scenario harnesses.

Brain2k-0005:
@codex review

Round 7 findings addressed and pushed: the boundary gate walks every scrollable ancestor from the touch target up to the panel (nested lists gate the drag), a mid-drag reversal into content-owned territory settles the panel back to its pre-gesture snap, rest offsets compose onto consumer-supplied inline offsets, the sticky handle suppresses native scrolling via touch-action none, and its background inherits from the panel so custom backgrounds show no seam.

Brain2k-0005:
@codex review

Note: the previous review round examined commit 4a933c1c, which predates the two most recent pushes — its findings on nested scroll containers, mid-drag reversal, offset composition, handle touch-action and handle background were already addressed in a1c8bd34. New since then: programmatic ActiveSnapPoint changes now route through the same disable-transition, convert, reflow, re-arm sequence as gesture settling, so a resting drawer animates smoothly from its current position instead of jumping (verified via getAnimations diagnostics from a non-zero snap). Current tip: 9c8834ce.

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

PR #382, opened 2026-07-23
fix(gantt): close the final review round's findings post-merge

Brain2k-0005:
PR #379's final review round attached three findings that were misread as clean and merged. This is the post-merge follow-up on master.

## Findings & fixes

**1. (P1) Navigation must supersede a suspended reconcile** — `Gantt3`
The prev/next/Today nav actions mutate `VisibleRange` directly, outside the reconcile pipeline, without claiming a generation. A reconcile suspended in its live-center capture await (a mode/geometry switch) could resume, still pass its `IsCurrentReconcile` check, and commit its now-stale mode/range on top of the navigation — the viewport would jump back to where the switch had been reshaping toward, discarding the nav the user just performed.
Fix: navigation claims the next generation via `SupersedeReconcileForNavigation()` — the same generation token the parameter and theme paths already guard on — so the suspended reconcile finds itself superseded and abandons its entire commit.

**2. (P2) Standalone timeline skips the duration filter** — `GanttTimeline`
`BuildFallbackRows` (the `Rows == null` path) copied every task 1:1, bypassing the duration filter every other GanttV3 entry point applies. A standalone consumer's invalid `End < Start`, non-milestone task rendered an 8px sliver bar instead of being dropped.
Fix: route `BuildFallbackRows` through `GanttRowModel.FilterValidDurationTasks` — the shared filter the `Rows`-fed path already uses. Milestones stay exempt (their `End` is forced to `Start` upstream).

**3. (P2) Shipped CSS bundles missing the group-header utility**
The prebuilt utility bundles were stale and missing the group-header stripe's arbitrary accent-opacity utility (`bg-accent/[0.18]`, matching v2's 0.18 fill-opacity), so a package-only consumer rendered the header transparent. The class is a static literal the Tailwind content scanner already extracts from the scanned Razor source — the bundles simply had not been rebuilt since it was added.
Fix: regenerated both bundles (root `lumeo-utilities.css` and docs `tailwind.out.css`) via the documented CSS pipeline. Each now carries the full-opacity fallback plus the `@supports` `color-mix` rule that resolves the header background to the accent color at 18% opacity. No safelist entry needed — reserved by convention for runtime-interpolated utilities, not statically-visible literals.

## Tests

- `GanttV3CodexRound18Tests` — nav (Today and prev/next) during a suspended reconcile supersedes it (gated live-center capture); standalone timeline drops an invalid-duration task and keeps a milestone.
- CSS fix verified by the pipeline: both bundles contain the escaped selector and the `var(--color-accent) 18%` `color-mix` rule; the E2E ServerHost consumes the rebuilt `_content/Lumeo/css/lumeo-utilities.css` and its group-header visual snapshots pass byte-unchanged.

## Gates

- Release `-warnaserror`: 0 warnings / 0 errors, both TFMs.
- Focused `~Gantt` unit filter: 300/300 both TFMs (296 prior unchanged + 4 new).
- Full Gantt E2E (ServerHost, Development): 60/60, all 8 visual snapshots byte-unchanged.
- Registry regenerated as the last step (one coverage association).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Gantt view-mode handling during navigation, theme changes, parameter updates, and delayed callbacks.
  * Prevented stale updates from overriding newer user selections.
  * Improved behavior when the component is disposed during asynchronous operations.
  * Corrected timeline rendering to exclude invalid-duration tasks while preserving milestones.
  * Prevented unnecessary task reprocessing and extra scroll actions.

* **Tests**
  * Added comprehensive coverage for view-mode ownership, replay behavior, navigation, disposal, and timeline filtering.

* **Chores**
  * Regenerated stylesheets and refreshed generated registry metadata.


Brain2k-0005:
@codex review

This follow-up closes the three findings from the final review round of the predecessor PR: navigation actions now supersede any suspended reconcile via the generation mechanism, the standalone timeline fallback routes through the shared duration filter, and the prebuilt utility CSS bundles are regenerated so the group-header background utility ships (verified stale rather than unscanned — no safelist entry needed).

Brain2k-0005:
@codex review

Round 1 findings addressed and pushed (current tip 8a721f74): superseded reconciles no longer lose their pending inputs. The mechanism now tracks a committed parameter reference alongside the eager one — a supersession leaves the committed reference behind, so the delta stays visibly pending, and the superseding action (navigation or theme change) re-runs the reconcile against current parameters after winning, landing the pending mode and task changes. Disposal still only supersedes. All prior supersession and matrix tests unchanged; new scenarios cover nav-during-suspended-mode-change, combined tasks-plus-mode supersession, and the theme path.

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

PR #386, opened 2026-08-10
fix: shadcn alignment wave 0 — dead classes, Card/Avatar rungs, Tabs pill geometry

Brain2k-0005:
## Summary

Wave 0 of the shadcn-alignment audit — cheap, non-breaking cleanups that clear the deck before the S1/S2 visual-breaking release (border opacity + radius rebase). Every item below is verified against the current shadcn `new-york-v4` registry, not memory.

## Changes

| # | Component | Change | Visible? |
|---|---|---|---|
| 1 | `Button.razor` | Delete dead Sm `text-xs` (`Cx.Merge`/TailwindMerge already discards it) | No — proven byte-identical |
| 2 | `UploadTrigger.razor`, `ShimmerButton.razor` | Delete the *same* `text-xs`, which their merge order let survive | Yes — 12px → 14px, matches Button + shadcn |
| 3 | `ThemeBuilder.cs` | Fix XML doc: wrong default (0.5 → 0.75) and wrong derivation (subtractive → multiplicative) | No |
| 4 | `QRCode.razor` | Delete inert `border-radius:4px` on an SVG `<image>` (no effect without `clipPath`) | No |
| 5 | `ImageCompare.razor` | Badge inline `border-radius:4px` → `rounded-sm` (6px token) | Yes, small — 4px → 6px |
| 6 | `Avatar.razor` | `AvatarShape.Themed` multiplier `*3` → `*4` | Fixes a shape bug live today under `.style-new-york` |
| 7 | `Card.razor`, `SidebarMenuButton.razor` | `rounded-lg` → `rounded-xl` / `rounded-md` (verified against shadcn registry) | Yes, 1px |
| 8 | `TabsTrigger.razor` | Pill padding `py-1.5`→`py-1`, add `dark:bg-input/30`, add `hover:text-foreground` | Yes, subtle — fixes a real track-overflow defect |
| 9 | `CustomizerSidebar.razor` | Fix misleading "New York" style description | No |

Not in scope (deliberately): radius-token rebase, border-opacity, focus-ring convention (Waves 1–3), and nothing under `src/Lumeo.Gantt/`.

### Item 6 detail — Avatar Themed circle math
[...]

PR #388, opened 2026-08-10
feat(size-scale): extend Lumeo.Size to all 7 rungs across 14 components

Brain2k-0005:
## Summary

Merges the finished `Lumeo.Size` batches into one branch and does the consolidated Tailwind CSS work they were deliberately withheld from committing individually.

Every component taking `Lumeo.Size` now implements all 7 rungs (`Xxs`/`Xs`/`Sm`/`Md`/`Lg`/`Xl`/`Xxl`) instead of silently falling back to `Md` for the unimplemented ones:

- **Batch A (partial)** — Avatar, Icon, Kbd, Rating. `Input` is intentionally excluded from this batch — it lives on the still-open `feat/shadcn-wave0` (PR #386) and follows in a later round to avoid a guaranteed merge conflict.
- **Batch B** — Switch, Toggle, ToggleGroup, ToggleGroupItem
- **Batch C** — SparkCard, Statistic, ReasoningDisplay, Chip, Spinner
- **Batch D** — Alert, List, ListItem, Result, FileUpload

**Why**: the `Size` parameter advertised a 7-value enum but components silently mapped unimplemented rungs to `Md` via `_ => ...` fallthrough. `Size.cs` requires components to document their supported set; none did until now.

**Purely additive, with one deliberate exception**: existing Sm/Md/Lg rungs are byte-identical to before — no visual change on upgrade for anyone already using those three. The one exception is **Avatar's `Size.Xs`**, called out below.

## Notable side effects found while doing this

- **Avatar's `Size.Xs` — fixed, deliberate visual change.** `Xs` previously fell through to the `Md` catch-all and rendered at 40px, *larger* than `Sm`'s 32px — a real inversion in shipped code. `Xs` now gets its own rung (`h-6 w-6`, 24px), restoring the correct visual order (`Xxs` 20px < `Xs` 24px < `Sm` 32px < ...). **Any consumer currently relying on `Size.Xs` on `Avatar` will see the avatar shrink** from 40px to 24px on upgrade.
- **Avatar's `Themed`-shape radius multiplier — fixed.** The corner-radius multiplier for `AvatarShape.Themed` (`rounded-[calc(var(--radius)*N)]`) was calibrated at `*3` to guarantee a full circle at stock theme radii "up to `Xl`" (3 × 0.75rem = 36px ≥ `Xl`'s 32px half-height). The new `Xxl` rung (80px, 40px half-height) broke that guarantee — 36px < 40px would have visibly squared the corners. Bumped to `*4` (48px ≥ 40px), which still covers every smaller rung too. Classified as a correctness fix, not cosmetic.
- **FileUpload's `Size` only affects `Variant="Button"`.** The Dropzone and Avatar variants were deliberately deferred; their XML docs now say so explicitly.
- **Touch targets**: several rungs fall under the 24px minimum, including some that already did before this change. **Rating is the worst offender in the whole spec** — the star buttons carry zero padding, so hit-area equals icon size exactly: `Xxs` 8px, `Xs` 12px, `Sm` 16px (pre-existing), `Md` 20px (pre-existing) all fail the 24px minimum; `Lg` 32px, `Xl` 36px, `Xxl` 40px pass. Not unilaterally fixed with padding (that's a layout change, the owner's call) — proposed as a follow-up: `p-2`/`p-1.5`/`p-1`/`p-0.5` at `Xxs`/`Xs`/`Sm`/`Md` respectively would each lift hit-area to exactly 24px without changing the visible star size. Switch also fails at `Sm` (16px track) **and `Md`, its default** (20px track) — pre-existing, unrelated to this PR.
- Four docs pages drove their Size picker with `Value="Small"/"Medium"/"Large"` against enum members actually named `Sm`/`Md`/`Lg`, so `Enum.TryParse` silently failed and the picker always rendered `Md` regardless of selection. `ReasoningDisplayPage`, `SparkCardPage`, and `StatisticPage` were already fixed in Batch C. **`RatingPage` is now fixed too** (`SegmentedItem Value="Small"` → `Value="Sm"`, `Value="Large"` → `Value="Lg"`; human-readable labels unchanged) — this was the last one.

## What this PR does (integration + the remaining Batch A component work)

- Merges `feat/size-scale-b` → `feat/size-scale-c` → `feat/size-scale-d` in order, then adds Avatar/Icon/Kbd/Rating directly on top (single-writer, no further merge). Component source files were disjoint by construction; the only merge conflicts were in generated registry/MCP catalogue files (`src/Lumeo/registry/*.json`, `docs/Lumeo.Docs/wwwroot/registry.json` + `registry/*.json`, `tools/lumeo-mcp/src/{registry,components-api}.json`), resolved by taking either side and then regenerating from source via `Lumeo.RegistryGen`.
- Icon and Kbd extended per spec (Icon already had `Xs`/`Sm`/`Md`/`Lg`/`Xl`; only `Xxs`/`Xxl` were missing). Kbd's height ties at `Sm`=`Md`=`h-5`, so the new small rungs use a flat 1-step Tailwind decrement rather than mechanical extrapolation (which would have frozen them at 20px too).
- Tailwind rebuild of `src/Lumeo/wwwroot/css/lumeo-utilities.css` and `docs/Lumeo.Docs/wwwroot/css/tailwind.out.css`, covering every utility class referenced across all batches, including this one's new Rating star-icon variants (`[&_svg]:h-2/w-2`, `h-3/w-3`, `h-9/w-9`, `h-10/w-10`). Verified via an exact literal-substring scan in Python (grep's escaping was unreliable on these minified single-line bundles and returned contradictory counts on this exact question in an earlier pass).
- No version bump (per repo convention: this is a code change pending a separate owner-approved tag).

## Test plan

- [x] Registry drift check (mirrors `.github/workflows/ci.yml`'s "Registry is up to date" step): empty diff.
- [x] Conflict-marker sweep + "adepto" sweep: clean.
- [x] Full solution build, `-warnaserror -m:1` (Debug and Release): 0 errors, 0 warnings (cross-checked against the build's own summary and a broad `: (error|warning) [A-Za-z]+[0-9]+:` regex over the log).
- [x] `scripts/docs-parity/check.py`: exit 0 (one pre-existing, unrelated `QRCodePage` casing warning, non-blocking).
- [x] Full filtered suite (`FullyQualifiedName!~Lumeo.Tests.E2E&Category!=EjectGateFull`): `Lumeo.Tests.dll` = 6821/6821 passed, all projects combined = 7116/7116 passed. Matches baseline (6788 / 7083) + 33 new tests added in this batch (4 `*SizeScaleTests.cs` files pinning all 7 rungs per component with exact space-delimited class-token assertions, plus 1 new Avatar `Themed`/`Xxl` radius regression test).
- [x] Per-component disable-check: for each of Avatar/Icon/Kbd/Rating, one new rung's switch case was temporarily commented out and the test suite re-run to confirm it fails with the concretely predicted fallback value (the `Md` catch-all in every case) before being restored — e.g. disabling Avatar's `Xs` case reproduces the original 40px inversion bug exactly, disabling Rating's `Xxl` case falls back to `[&_svg]:h-5 [&_svg]:w-5`, etc.

---

## Update: Codex review findings closed, scale-ordering test added, touch targets patched

Codex reviewed this PR and returned 2 P2 findings, both fixed, plus the structural gap they exposed (a scale-ordering test), plus a separate owner-approved touch-target follow-up. All pushed to this branch.

### Finding 1 — Switch's Loading spinner didn't scale (fixed)

`Switch` passed a **fixed** `Size="Sm"` to the nested `<Spinner>` and tried to override the visible size via a wrapper `Class="!h-X !w-X"`. `Spinner` puts its size classes on its own inner `<svg>` (never on the wrapper `<div>` that `Class` merges into), so the override never reached the ring — it rendered a constant 16px `<svg>` at **every** Switch rung, overflowing the 4–8px `Xxs`/`Xs` thumbs and never growing past `Sm` at `Xl`/`Xxl`.

Fixed by mapping Switch's `Size` to Spinner's own `Size` parameter (`SpinnerSizeForThumb`), so the real `<svg>` resizes. Spinner only exposes 7 discrete rungs that don't line up 1:1 with the thumb's sizes, so `Lg` rounds down and `Xxl` rounds up (documented in the component) — the ring still strictly grows across the top of the scale instead of tying at a fixed size.

**Wrapper-vs-inner sweep** (same bug shape, across the rest of this PR's components): found one more instance — `ReasoningDisplay`'s chevron hardcoded **both** its own wrapper size (`h-3.5 w-3.5`) **and** a fixed nested `<Icon Size="Sm">`, so it never tracked the component's own `Size` at all (a constant ~14px chevron next to 9px text at `Xxs` and 18px text at `Xxl`). Fixed the same way — `Size="@Size"` passed through to the icon, reusing Icon's own `Xxs..Xxl` scale (same one Alert already uses). No other components in the PR had this shape; Alert/Chip/Result/Rating/Kbd/Statistic all size their nested `<SvgGlyph>`/icon directly with no wrapper indirection.

### Finding 2 — FileUpload's `Xl`/`Xxl` padding regressed below `Lg` (fixed)

`Lg` was `px-8` while `Xl`/`Xxl` were `px-5`/`px-6` — narrower than `Lg` despite height/font-size growing. Raised to `px-9`/`px-10`; every other rung (including `Lg`, which is FileUpload's own historical value, deliberately not aligned with Button's `lg`) is untouched.

### Structural gap — scale-ordering test

Added `tests/Lumeo.Tests/Contract/SizeScaleOrderingTests.cs` (60 test cases) plus a shared Tailwind-class-to-pixel parsing helper (`tests/Lumeo.Tests/Helpers/SizeScaleAssert.cs`). It renders every Size-driven component in this PR at all 7 rungs in **visual** order (`Xxs < Xs < Sm < Md < Lg < Xl < Xxl` — not enum declaration order, where `Xxs = 6` is declared last) and asserts each dimension is monotonically non-decreasing.

Two more pre-existing, undocumented ties surfaced: `Alert.DescriptionClass` and `Chip.SizeClass` both tie `Md`/`Lg` at `text-sm` instead of stepping up the way their sibling dimensions (`Alert.TitleClass`, `Chip`'s own radius) do. Verified via `git show origin/master:...` that **both predate this PR/campaign** — so, per the "never change an already-implemented rung" rule already established for `Chip.CloseIconSize`'s identical gap, they're left as-is and now documented explicitly with dedicated tie tests (not just silently tolerated by the non-strict `>=` ordering check).

Disable-check: reverted the FileUpload padding fix and confirmed `SizeScaleOrderingTests.FileUpload_Button_Padding_Is_Monotonic` fails with the exact predicted message (`Xl (20px) < Lg (32px)`) before restoring it.

### Touch-target follow-up (separate owner decision, same branch)

`Rating`'s star buttons and `Switch`'s track both have zero padding, so hit-area equals visible size — several rungs sit under the 24px touch-target minimum. The owner chose to patch **only `Sm` and `Md`** (the two rungs reached without a deliberate dense-size opt-in — `Md` is the default), leaving `Xxs`/`Xs` deliberately tiny (picking them is itself a request for a dense control) and `Lg`+ untouched (already ≥24px).

- **Rating**: real button padding — `p-1` at `Sm` (16px icon + 2×4px = 24px exact), `p-0.5` at `Md` (20px icon + 2×2px = 24px exact). The icon's own `[&_svg]:h-N/w-N` token is untouched, so the **visible star glyph is pixel-identical** to before — only the button's hit box grows. Side effect (honest disclosure): since the container's `gap-0.5` between star buttons is unchanged, the growth in each button's box widens the *visible* edge-to-edge gap between adjacent stars: `Sm` 2px → 10px, `Md` 2px → 6px. Overall widget width (5 stars, default `Max`) grows `Sm` 88px → 128px (+45%), `Md` 108px → 128px (+19%).
- **Switch**: an invisible `::before` hit-zone (`-inset-y-1` at `Sm`, `-inset-y-0.5` at `Md`) rather than padding, because the track *is* the painted `<button>` (background lives on the button itself) — real padding would visibly enlarge the pill, not just its hit box. Mirrors the same pattern `Chip`'s close button already uses (`before:absolute before:-inset-2`). `-inset-y` only, not `-inset-x`: track width already clears 24px at every rung. **Visible track and outer layout are pixel-identical** to before (the pseudo-element is absolutely positioned, out of flow — no gap/spacing side effects at all, unlike Rating). Trade-off: the invisible zone extends 4px (`Sm`) / 2px (`Md`) above and below the visible track and could overlap very tightly-stacked neighboring content.

Both components' `Size` XML docs now state the `Xxs`/`Xs` exception explicitly. Disable-check: reverted both fixes and confirmed all 8 new touch-target tests fail with the exact predicted pre-fix values (16px/20px instead of 24px) before restoring.

### Gate (updated)

- Conflict-marker + "adepto" sweep: clean.
- Full solution build, `-warnaserror -m:1` (clean obj/bin): 0 errors, 0 warnings (Debug and Release).
- Registry/MCP catalogue regenerated from source and re-synced (`docs/Lumeo.Docs/wwwroot/registry.json`, `registry-search.json`) — diff against CI's own drift-check command contains only the expected content changes (test counts / tier / relatedFiles / behavior / a11y flags reflecting the real source edits) plus the excluded `"generated"` timestamp.
- `scripts/docs-parity/check.py`: exit 0 (same pre-existing, unrelated `QRCodePage` casing warning as before).
- Full filtered suite (`FullyQualifiedName!~Lumeo.Tests.E2E&Category!=EjectGateFull`): `Lumeo.Tests.dll` = **6914/6914** passed (net8.0 and net10.0 lanes both green), all projects combined = **7209/7209** passed. Baseline was 6821 / 7116 → **+93 tests**, fully accounted for: 60 new (`SizeScaleOrderingTests`) + 16 (Switch touch-target + finding-1 regression tests) + 9 (Rating touch-target) + 7 (ReasoningDisplay chevron) + 1 (Alert tie fact) = 93.
- CSS bundles rebuilt for the new utilities this round introduced (`px-9`, `px-10`, `p-1`, `p-0.5`, `before:-inset-y-1`, `before:-inset-y-0.5`, `before:inset-x-0`) — verified present in both `src/Lumeo/wwwroot/css/lumeo-utilities.css` and `docs/Lumeo.Docs/wwwroot/css/tailwind.out.css` via a literal substring scan in Python.
- build-server shutdown, no stray `obj_*`/logs/debug artifacts committed.

No version bump (per repo convention — pending a separate owner-approved tag).

## Summary by CodeRabbit

* **New Features**
  * Expanded component sizing support across the UI library, covering seven size options from Xxs through Xxl.
  * Added size-specific styling for spacing, typography, icons, controls, avatars, and loading indicators.
  * Updated interactive documentation and examples to showcase the complete size range.
  * Added guidance for small-size touch-target and rendering behavior.

* **Bug Fixes**
  * Corrected size rendering for avatars, alerts, chips, and other components at previously unsupported sizes.

* **Tests**
  * Added comprehensive size-scale coverage across 16 components.


Brain2k-0005:
@codex review

Brain2k-0005:
@codex review

PR #392, opened 2026-08-10
refactor(charts): trim native rendering engine to a lightweight sparkline core

Brain2k-0005:
## Summary

**This PR no longer replaces ECharts.** The owner reversed that decision: ECharts stays as the rendering engine for all 30 chart types, and the native engine here is demoted to what it's actually good for — **lightweight cases only** (sparklines, mini charts inside `DataGrid` cells and similar), where pulling in a ~1 MB charting engine for a 64×24px trend line isn't justifiable.

This update trims the branch accordingly.

## What was removed, and why

The **14 native chart types** (`NativeLineChart`, `NativeAreaChart`, `NativeBarChart`, `NativeMixedChart`, `NativeScatterChart`, `NativeEffectScatterChart`, `NativeWaterfallChart`, `NativePieChart`, `NativeDonutChart`, `NativeNightingaleChart`, `NativeRadarChart`, `NativeHeatmapChart`, `NativeBoxPlotChart`, `NativeCandlestickChart`) are gone, along with:

- Their hosts, `CartesianChartHost`/`XyChartHost`.
- The full-chart chrome components that existed only to serve them: `ChartAxis`, `ChartLegend`, `ChartCrosshair`, `ChartZoomSlider`, `ChartCanvasSurface` (and the Canvas-fallback draw-command interop path it needed — `ChartCanvasDraw`, `ChartCanvasCommand`).
- Their bUnit tests, and the `/e2e/charts-native` side-by-side ECharts-vs-native comparison page + its smoke test.

They were built for a goal the owner is no longer pursuing — keeping them around would mean maintaining 14 components nothing renders. Git history retains them if they're ever wanted back.

## What survives — the sparkline core

The pure, well-tested foundation stays, because it's exactly what a sparkline needs:

- **Scales**: linear, log, time, band, point + `NiceTicks`/`TimeTicks`.
- **Geometry**: `ChartPathBuilder`, `ChartArcPath`, `ChartRoundedRectPath`.
- **Downsampling**: `Lttb`/`ChartDownsampling`.
- **Color**: `ChartColorScale`.
- **Layout**: `ChartLayout`, the coordinate systems, `ChartRenderStrategy`.
- **The aspect-ratio fix**: `NativeChartScreenTransform` + the `ChartObserveBox`/`ChartUnobserveBox` `ResizeObserver` interop — this is what fixes distorted circles and stretched text, and it matters just as much at 64×24px as it does on a full chart.
- **Hover/tooltip primitives a sparkline would plausibly use**: `ChartInteractionSurface` (pointer hit-testing + keyboard nav) and `ChartTooltipHost`, plus their `Interaction`-layer support (`ChartHitTester`, `ChartHoverState`, `ChartKeyboardNav`).

Boundary calls (per-component reasoning, not mechanical deletion):
- **Kept** `ChartInteractionSurface`/`ChartTooltipHost` — small, generic, and a DataGrid-cell sparkline plausibly wants hover-to-see-exact-value.
- **Dropped** `ChartAxis`/`ChartLegend`/`ChartCrosshair`/`ChartZoomSlider`/`ChartCanvasSurface` — full-chart chrome (visible axes, legends, brush/zoom UI, a dashed full-height crosshair guideline) that doesn't fit a mini chart's footprint or purpose.
- **Dropped** `ChartBrushState`/`ChartZoomState`/`ChartLegendState`/`ChartStatistics` (box-plot quartiles) — orphaned once their only consumers (zoom slider, legend, box plot) went away.

## Confirmed: the ECharts path is untouched

`Chart.razor`, `EChartOption.cs`, `echarts-interop.js`, and all 30 legacy chart wrappers are byte-for-byte identical to `master`:

```
git diff origin/master -- src/Lumeo.Charts/UI/Chart/Chart.razor src/Lumeo.Charts/UI/Chart/EChartOption.cs src/Lumeo.Charts/wwwroot/js/echarts-interop.js
git diff origin/master --stat -- src/Lumeo.Charts/UI/Chart/Charts/*.razor
git diff origin/master --stat -- src/Lumeo.Charts/UI/Chart/*.razor src/Lumeo.Charts/UI/Chart/*.cs
```
All three: empty diff.

## A `Sparkline` already exists — it doesn't use this core, and shouldn't be forced to

The docs nav already lists a `Sparkline` (`src/Lumeo/UI/Sparkline/Sparkline.razor`, part of the core `Lumeo` package). It renders line/area/bar sparklines via its own small, self-contained inline-SVG string builder with its own min/max scale math — no dependency on `Lumeo.Charts` at all.

`Lumeo.Charts` depends on `Lumeo` (not the other way around), so wiring the existing `Sparkline` to this trimmed core would mean either inverting that dependency or shipping it from a different package — a real architectural change, not a small one, for essentially the same behavior `Sparkline` already has (Line/Area/Bars, last-point highlight, native `<title>` tooltips, single-value/empty fallback). Not attempted here. The core exists for a **future** consumer (e.g. a `DataGrid` cell mini-chart) that actually needs its downsampling/interaction primitives — that consumer doesn't exist yet, and a core with no consumer will rot if one isn't built before too long.

## Test counts

- Before this trim (`master` merged in, native engine still full-size): **7886 passing**.
- After (native engine trimmed): **7690 passing**.
- Decrease of **196 tests**, entirely accounted for by the 33 deleted test files (3590 lines) covering the 14 deleted chart types, their hosts, and the dropped chrome components. This is expected and correct, not a regression — a test suite for deleted code is noise.

## Gate results

- [x] Registry drift (CI's own regenerate-and-diff) — empty.
- [x] Conflict-marker + "adepto" sweep — clean.
- [x] `dotnet build Lumeo.slnx -warnaserror -m:1` after nuking obj/bin — 0 warnings, 0 errors (cross-checked with a broad `: (error|warning) [A-Za-z]+[0-9]+:` regex over the raw log — 0 matches).
- [x] `python scripts/docs-parity/check.py` — exit 0. Removed the now-obsolete `NativeBoxPlotChart.Samples` allowlist entry (its type no longer exists).
- [x] Full filtered suite (`Lumeo.Tests`, `-f net10.0`, E2E excluded): **7690/7690 passing**.
- [x] Local E2E on an isolated port (docs host on `127.0.0.1:58231`, never touching the owner's `5287`): **31/31 smoke tests passing** (a `/components` catalog render check, DataGrid, dialogs, dropdowns, popovers, search palette, select, slider, tooltip). Confirmed the deleted `/e2e/charts-native` page has no remaining references anywhere in the repo — including the Playwright regression test that targeted it (`AspectRatioDistortionTests`), which is removed along with its subject.
- [x] No stray `obj_*`/logs committed; build server and the isolated test server both shut down by PID (never touched port 5287).


Brain2k-0005:
Closing without merging.

The native rendering core was built for the ECharts replacement, which the owner reversed after the side-by-side comparison page showed how much rendering polish a decade of ECharts represents. The fallback justification — that the core would carry sparklines — did not survive checking: a self-contained `Sparkline` already ships in the core `Lumeo` package with no `Lumeo.Charts` dependency, and since `Lumeo.Charts` depends on `Lumeo` rather than the reverse, wiring the two together would mean inverting that dependency for essentially the behaviour it already has.

So the core has no consumer. Merging it would add maintained code that nothing renders.

The work is not lost — scales, tick selection, LTTB downsampling, path building, layout and the ResizeObserver-based aspect-ratio fix all remain in this branch's history and can be cherry-picked if a real need appears (a richer inline chart in DataGrid would be the obvious one).

What this effort did produce, and what stays: the side-by-side comparison page surfaced nine rendering defects that ~7,900 passing tests never caught — including a heatmap rendering every cell one colour and hover resolving the wrong category — and it gave the owner something real to judge, which is what made the decision to keep ECharts possible.

PR #394, opened 2026-08-10
feat(charts): give the ECharts theme layer real design polish

Brain2k-0005:
## Summary

Fresh direction (decided after comparing the native engine against ECharts): the engine stays, the design layer is the work. This makes the ECharts theme layer — which already existed as the single styling seam (`registerLumeoTheme` in `echarts-interop.js`) — into the shared design baseline the 30 chart wrappers inherit, rather than adding a parallel mechanism.

- **Legend**: centred (`left: 'center'`) via the theme only. Reaches all 22 legend-using wrappers with zero wrapper edits — ECharts fills an unset option property from the theme, and none of them set `Left` themselves.
- **Tooltip**: rebuilt to match the Lumeo `<PopoverContent>` surface (`bg-popover`/`text-popover-foreground` equivalents, border, `var(--radius-md)`, `var(--shadow-lg)`) and given a real fade (`transitionDuration` + an explicit `opacity`/`transform` CSS transition on the tooltip DOM node) instead of a snap.
- **Hover**: `emphasis.focus` + a matching `blur` state added per series type, so hovering one series visibly dims the rest instead of leaving everything at full opacity.
- **Animation**: explicit, distinct entrance vs. update durations/easing. `prefers-reduced-motion` is checked fresh on every theme (re)registration and enforced as a hard `animation:false` override in `initChart`/`updateChart`, so it always wins over a consumer's own `AnimationDuration`/`AnimationEasing`.
- **Ticks**: value axes get `minorTick`/`minorSplitLine` for materially finer grid density, entirely from the theme — `EChartAxis` has no `SplitNumber` property today, so no wrapper was requesting a count either way.

### Bug found and fixed while verifying theme-switch propagation

Verifying "does a dark-mode/scheme switch actually reach an already-rendered chart" with computed values (not a screenshot) turned up a real pre-existing bug: `refreshAllCharts()` rebuilt the re-themed chart instance from `chart.getOption()`, which ECharts always returns fully resolved — every themed colour comes back baked in as a concrete hex value. Re-feeding that into a freshly re-themed instance meant the OLD theme's colours silently shadowed the NEW theme's; a dark-mode toggle was a no-op on any chart already on screen. Confirmed live: series colour stayed pinned at `#18181b` after toggling dark + firing `lumeo:theme-changed`, even though `--color-chart-1` itself had already updated.

Fixed by stashing the pristine pre-resolution option JSON per chart instance and replaying it through a fresh CSS-var resolution pass into the freshly re-themed instance. Verified fixed live: chart-1 now resolves to `#d9d9d9` in dark mode and `#2463eb` under the `blue` scheme.

### Consolidation

The `if (AnimationDuration.HasValue) ... if (!string.IsNullOrEmpty(AnimationEasing)) ...` pair was duplicated verbatim across 28 of the 30 chart wrappers — one design decision copy-pasted 28 times. Consolidated into `ChartHelper.ApplyAnimation(option, duration, easing)`.

No public component parameters changed — presentation only.

## Test plan

- [x] `node --test tests/js/echarts-interop-theme.test.mjs` — 11 new tests asserting computed theme values (legend.left, minorTick.splitNumber, tooltip.transitionDuration, exact `extraCssText` fragments, animation durations/easing, per-series-type emphasis/blur, the chart-1-vs-primary contrast trap, `applyReducedMotion`'s hard-override behaviour). Each disabled-in-place once to confirm the predicted wrong value, then restored.
- [x] Full `dotnet build -warnaserror` — 0 errors/warnings (cross-checked against a broad regex over the log).
- [x] Registry regen + docs-parity check — exit 0.
- [x] `Lumeo.Tests` full suite — 7301/7301 passing (chart-specific subset: 130/130).
- [x] `Lumeo.Tests.E2E` full suite (excluding the two documented pre-existing flaky classes) — 190/190 passing, run against isolated local servers.
- [x] Live Playwright verification (screenshots + computed ECharts instance state) across `/components/chart` and 6 individual chart pages, in light/dark and two named schemes (zinc default, blue, violet) — zero console/page errors.

### Known remaining weakness

The tooltip's fade-in/out is fully covered by computed-value tests on its *configuration* (transitionDuration, the CSS transition string), but I was not able to reliably capture the live hover state in headless Chromium within the time available — this is genuinely judged by eye and I did not fake coverage of it with a test that doesn't actually exercise motion.

---

## Follow-up pass: gradients, glows, staggered reveals ("distinctive, not just correct")

The first pass above made the theme layer correct and consistent — legend, tooltip, hover, ticks, the theme-switch bug. It did not make the charts look designed; every chart still read as "ECharts with a palette applied." This pass adds the four devices the brief actually asked for, then deliberately does **not** use all four everywhere.

### Gradients — per series, not a flat literal

`itemStyle.color` on a theme series-type block can be a callback `(params) => Color` — confirmed with a live probe against real ECharts, registering a theme with callback-valued `bar.itemStyle.color` / `pie.itemStyle.color` / `line.lineStyle.color` / `line.areaStyle.color` and checking which ones actually fired. Only `itemStyle.color` (bar, pie, and by extension every other item-based series type) invokes the callback; `lineStyle.color`, `areaStyle.color`, and `shadowColor` on any style block never do, on any series type. That result shaped every gradient/glow decision below — it isn't a stylistic choice, it's what ECharts will actually let a shared theme do.

- **Bar** (`bar.itemStyle.color`): a vertical gradient built from `params.color` — the series' own already-resolved colour (theme token or a consumer's custom `Colors`/`ColorPalette`, doesn't matter which). Top stop is that colour lightened 32% toward white, bottom stop is the colour at full strength. Reads as "lit from above, grounded at the base" — deliberately **not** a fade to transparent; a bar sitting on a card needs a solid foot.
- **Pie/donut/nightingale** (`pie.itemStyle.color`, same series type under the hood): a **radial** gradient per slice, lightened 28% at the centre fading to the full colour at the outer edge — a "glassy" pop that reads well on both the flat-teal light theme and the near-monochrome dark palette this repo's chart tokens currently use.
- **Line stroke**: stays a **solid token colour**, not a gradient. Two reasons, not one: legend swatch and stroke need to read as the same colour at a glance, and the callback finding above means a per-series line gradient isn't reliably achievable at this seam anyway (only itemStyle.color fires).
- **Area fill** (AreaChart): the wrapper already had a real per-series `BuildGradient()` helper (top stop = the series' resolved colour, bottom = transparent) sitting behind an opt-in `GradientFill` flag that nobody used by default. The *default* fill (`Gradient=true`, `GradientFill=false` — unchanged parameter names/defaults) now routes through that same helper at a moderate 0.45 top opacity instead of the old flat 0.3-opacity rectangle. `Gradient=false` keeps the flat, denser, no-gradient look unchanged — a real "solid" alternative still exists. The explicit `GradientFill`/`GradientStops` opt-in (full-opacity top, the boldest "brand" look) is untouched.

### Glows — brand-tinted, not per-series, and that's a direct consequence of the finding above

Since `shadowColor` never accepts a callback, a theme-level glow is necessarily **one literal colour shared by every series of that type** — there is no way to tint it per series from this seam. Rather than fight that with a neutral grey/black shadow, the glow uses `--color-primary` (the app's own accent, resolved once per theme build) at low alpha. Every glow across every chart reads as "this app's accent," not as a generic drop-shadow:
- **Line**: `shadowBlur: 6` / 16% alpha at rest, `shadowBlur: 14` / 34% alpha on emphasis — subtle ambient lift normally, materially stronger under the pointer.
- **Bar/pie/scatter/heatmap**: emphasis-only (retinted from the previous pass's neutral `rgba(0,0,0,…)` shadows to the same brand-tinted alpha) — a resting glow on every bar in a dense chart would be noise, so it's reserved for "which one am I hovering."
- **Scatter** is the one exception that also glows at rest (`shadowBlur: 4`, 18% alpha) — sparse individual points, unlike a dense bar/pie fill, so a light halo earns its place without adding visual noise.

### Reveal — staggered per family, capped, and gated hard on reduced motion

- **Bar**: `animationDelay: idx => min(idx * 45, 700)` — a left-to-right cascade instead of every bar growing from the baseline simultaneously, capped so a 30-category chart still finishes in well under a second.
- **Pie/donut**: `animationDelay: idx => min(idx * 90, 500)` — slices sweep in around the circle in order instead of expanding together.
- **Scatter**: `animationDelay: idx => min(idx * 10, 400)` — fast cascading pop-in, capped for dense point clouds.
- **Line/area**: no per-point delay (the stroke/area is one continuous shape, not per-datum elements — `animationDelay` has nothing meaningful to stagger). Instead it gets its own longer, more deliberate entrance duration (900ms vs. the shared 700ms default) layered on top of ECharts' native left-to-right clip-path draw, so it reads as "the line is drawing itself in" rather than "the chart popped up."
- All of the above is **omitted outright** (not merely inert) when `prefers-reduced-motion` is active, via the same spread-conditional pattern the first pass used for the root animation block — verified with a computed-pixel-growth test (see below), not just "the code looks right."

### Held back: decal/pattern textures

This is the one device from the brief's list of four that this pass does **not** ship, and the reasoning matters as much as the other three:

The shared theme has no visibility into how many data points a given chart instance will render — it's a static object merged in before any data exists. A theme-level `pie.itemStyle.decal` would apply identically to a 3-slice donut and a 12-slice nightingale; the brief's own guidance is that decals are "noise on a dense line chart" and only "useful on a small set of categorical bars or pie slices" — i.e. the value of the device depends entirely on the thing the shared seam can't see. The one place a decal would be safely scoped (stacked-bar segments, where segment count is bounded by series count, not data count) would require per-wrapper edits to `BarChart.razor`'s existing `Stacked` branch, which was out of budget for this pass alongside verifying the other three devices properly. Chose not to guess at a weakly-scoped theme-wide decal rather than ship a fourth effect that reads as noise on half the chart pages.

### Theme-switch survival

Verified live (Playwright, real `themeManager.setMode`/`setScheme` calls with charts already rendered — not a reload) that every gradient/glow above survives a switch, in both directions:
- Light → dark → **violet** (dark) → light+violet, on the Area chart's default gradient fill, screenshotted after each switch. Gradient top/bottom stops re-resolve to the new scheme's `--color-chart-1`/`--color-chart-2` every time; nothing stays pinned to the pre-switch hex.
- Bar/pie/donut gradients and scatter/line glows spot-checked in light and dark on their own docs pages.
- This relies on the SAME pristine-JSON replay mechanism the first pass's bug-fix introduced (`refreshAllCharts` re-registers the theme and replays each chart's stashed pre-resolution option through fresh `resolveCssVars`) — the gradient/glow colours are built inside `buildLumeoTheme` itself, so a theme re-registration naturally rebuilds them from the live CSS vars at that moment. No separate mechanism was needed to make this survive; it was the direct benefit of building on the existing seam rather than a parallel one.

### Rigor

- Colour helpers (`hexToRgb`, `withAlpha`, `lighten`) are pure functions with exact computed-value tests (`rgb(197, 82, 82)` for `lighten('#aa0000', 0.32)`, not "some other colour came out").
- Bar/pie gradient callbacks tested by invoking them directly with a fixed `params.color` and asserting the exact returned gradient descriptor (type, coordinates, both stop colours) — not that "a color property exists."
- The animationDelay cap was disabled in place (`Math.min(..., 700)` → uncapped) to confirm the concretely predicted wrong value (`4500`, i.e. `100 * 45`) before restoring — a disable-check that couldn't have failed would have meant the fixture was wrong.
- Reduced-motion actually disabling the reveal (not just theoretically) was verified with a computed-pixel test: painted-pixel count in the bar chart's canvas over the first ~900ms after first paint grows 316% under normal motion (bars visibly still growing) and is flat (−0.0%) under `reducedMotion: 'reduce'` (bars render at final height immediately) — a real Chromium instance, not an assumption about what `animation:false` does.
- `Lumeo.Tests` full suite: 7302/7302 passing on both net8.0 and net10.0 (delta of +1 per framework from the 7301 baseline, fully accounted for by the one new xUnit fact this pass adds — `Default_Gradient_False_Is_Opacity_Only_No_Gradient`).
- `Lumeo.Tests.E2E`: 191/193 on the full run; the 2 failures are the documented pre-existing flakes (`HomePageVisualTests` — 8.43% Windows-local pixel diff vs. the Ubuntu baseline, never recaptured; a Gantt drag-parity test that passed cleanly when re-run in isolation, confirming the documented "fails only under heavy parallel load" class). Neither test touches charts.

### Still not fully there, in my own judgement

- The gradients and glows are real and verified, but on this repo's current default chart palette (which trends near-monochrome — black/white/teal rather than a wide hue spread) the radial pie gradient in particular reads more "glassy sphere" than "brand colour fade," because there isn't much colour for the lighten-toward-white math to work with. It looks distinctly more crafted than before, but a palette with more chroma spread across chart-1..5 would show the device off better than this repo's own docs demos currently do.
- Decal/pattern is a genuine gap against the brief's four devices, not a stylistic substitution — see "Held back" above.
- I did not add a bar-family texture/pattern anywhere, including the one place (stacked segments) where it would have been safely scoped; that's explicitly future work, not something this pass quietly declined to mention.

---

## Follow-up pass: differentiate by form, not colour

The owner decided to **keep shadcn's chart palette unchanged**. That's a real
constraint, not a stylistic preference: `--color-chart-1` is achromatic
(chroma ≈0 in both light and dark — the first series in every chart renders
grey), `chart-3` is weak in light, and `chart-4`/`chart-5` sit only ~29° apart
in hue. Colour cannot carry "which series is this" anymore. This pass makes
form do that job instead, for the two chart families where it earns its
place: categorical bar/pie, and multi-series line/area.

### Decal — now shipped, with the scoping the previous pass correctly held back on

The previous pass in this PR declined decal because the shared *theme* has no
visibility into how many data points a chart instance will render — a
theme-level `pie.itemStyle.decal` would hit a 3-slice donut and a 30-slice
sunburst identically. That reasoning was sound. The palette decision changes
the calculus (patterns are no longer decoration, they're the only remaining
way to tell two similarly-toned series apart), so this pass ships decal —
scoped **per wrapper**, where the actual item/series count is known:

- `ChartHelper.GetDecal(index, totalCount)` returns one of five hand-tuned
  decal recipes (diagonal stripes / dots / fine grid / diamonds / triangles —
  genuinely different shape families, not rotated variants of each other),
  cycling every 5 like the colour palette. Returns `null` — no decal emitted
  at all — when `totalCount < 2` (nothing to differentiate) or
  `totalCount > MaxDecalItems` (8): beyond 8 the 5-recipe cycle just repeats
  (same period as the colour palette, so nothing *new* is communicated) while
  making an already-thin pie sliver or a crowded bar cluster busier to read.
  This bounded-count scoping is what makes turning decal on at all safe — the
  "cardinality problem" the previous pass declined it over.
- **Wired into**: `BarChart` (grouped/stacked, ≥2 series — one decal per
  series, same as colour), `PieChart`/`DonutChart`/`NightingaleChart` (one
  decal per slice, ≤8 slices). Not wired into single-series bar/line — see
  "the achromatic-series question" below for why single-series gets a
  different treatment.
- **A live probe (registering a theme with a callback-valued
  `bar.itemStyle.decal`) confirmed `itemStyle.decal`, unlike `itemStyle.color`,
  is NOT invoked as a theme-level callback function** — `callbackHit` stayed
  `0` after a full render. That's why decal is assigned per-series/per-slice
  in the wrapper components (where the count is known) rather than at the
  shared theme seam the color gradients use.
- Decal colour is always `var(--color-border)` — a token, resolved by the
  existing `resolveCssVars` pass exactly like every other `var(...)` already
  flowing through a chart option. No raw hex.

### Line/area — marker symbol + dash + stroke weight cycle per series

A live probe confirmed ECharts does **not** cycle marker symbol by default
the way it cycles colour — three plain `line` series with no `symbol` set
all came back `emptyCircle` (`symbolSize: 4`) from `getOption()`. Unlike
decal, this doesn't need per-chart data to scope safely, so
`ChartHelper.ApplyLineForm(series, index)` is applied to every series in
`LineChart`/`AreaChart`:

- Symbol cycles `circle → diamond → triangle → rect → roundRect` (5, same
  period as the palette).
- Dash cycles `solid → dashed → dotted` starting at series 1 (series 0 stays
  solid).
- Series 0 gets `width: 3` instead of the theme's default `2` — see below.
- Verified merge behaviour is additive, not replacing: a per-series
  `lineStyle: { type, width }` merges over the theme's `lineStyle.shadowBlur`/
  `shadowColor` rather than clobbering them (checked via a live probe against
  the real theme + a partial per-series `lineStyle`).
- Hover emphasis now also enlarges the marker (`emphasis.scale: 1.5` on the
  `line` theme block) alongside the existing stroke-thickening, so "which
  series am I on" is legible by two independent form cues, not one.

### The achromatic-series question

The brief specifically asks: what should a *single-series* chart — the most
common case, and the one where chart-1's near-zero chroma is unmasked by no
other series being on screen — look like? Two different answers, deliberately
not the same lever everywhere:

- **Line/area**: series 0 gets a bolder stroke (`width: 3` vs. the theme
  default `2`). A single grey line reads as "the primary line, drawn with
  intent" on weight, rather than leaning on a hue it doesn't reliably have.
  Combined with the pre-existing brand-tinted glow (previous pass), this was
  judged sufficient — a decal running the length of an entire single stroke
  would be closer to noise than signal.
- **Bar/pie**: intentionally **no decal** for a lone series/slice — nothing to
  differentiate against, and the existing top-lightened / radial gradient
  (previous pass) already gives it a "lit, not flat" read. Texturing a chart
  that has nothing to distinguish would be decoration for its own sake, which
  the brief explicitly flags as the failure mode to avoid now.

### Verification

- `docs/Lumeo.Docs` on an isolated local port, screenshotted (Playwright
  driving the vendored Chromium directly — `agent-browser`'s daemon doesn't
  start on this machine) across `/components/charts/{bar,line,area,pie,
  donut,nightingale}` in light/dark × zinc/rose/violet, **before performing
  the theme switch on an already-rendered page** (matching the non-negotiable
  in the brief) — confirmed via the SAME `refreshAllCharts` pristine-JSON
  replay path the previous pass's bug-fix relies on.
- **Greyscale test**: one light-scheme and one dark-scheme screenshot of each
  family converted to true greyscale (`sharp .greyscale()`). Bar: two
  identically-toned grey bars remain tellable apart by stripe-vs-dot texture.
  Line: solid/thick vs. dashed vs. dotted remain tellable apart with zero hue
  information. Pie: all 5 slices remain tellable apart by pattern alone. This
  is the single clearest measure the brief asks for, and it passes.
- **True before/after**: stashed the change, rebuilt, and screenshotted the
  SAME bar/line/pie demo pages pre-change for a real diff (not a
  recollection) — confirms the predicted before-state: flat solid bars/pie
  slices differentiated only by hue, and (on this repo's own dark, low-chroma
  default palette) the "Basic Pie Chart" demo's 5 slices were close to a
  single near-black circle with thin dividers before this pass.
- Computed-value tests (`ChartFormDifferentiationTests.cs`, 12 new xUnit
  facts) assert the actual rendered `EChartOption.ToJson()` output — decal
  presence/absence at the count boundaries (1, 2, 8, 9 items), exact decal
  `symbol`/`color` values, exact per-series `symbol`/`lineStyle.type`/
  `lineStyle.width`, and that the stacked-bar corner-radius `ItemStyle`
  survives the decal being layered onto it. Each of the two wrapper-level
  loops (`BarChart`'s decal loop, `LineChart`'s form loop) was disabled in
  place and the run re-verified: the tests asserting *presence* failed with
  the concretely predicted `KeyNotFoundException`/message, the tests
  asserting *absence* (single-series, >8 items) were unaffected — then
  restored.
- A new JS test (`echarts-interop-theme.test.mjs`) asserts
  `theme.line.emphasis.scale === 1.5` as a computed value.

### Held back / not fully there

- Decal is scoped to bar/pie/donut/nightingale. Radar, sunburst, treemap,
  and the other categorical-ish series types were left alone — they either
  have an unbounded/hierarchical cardinality decal can't safely scope to, or
  their existing form (radar's polygon shape, treemap's rectangle packing)
  already carries enough distinction without it.
- A dense multi-line chart (10+ series) was deliberately NOT given decal or
  an expanded dash vocabulary — past ~5 series the symbol/dash cycle repeats
  exactly like the colour palette does, which is consistent, not a gap, but
  it's worth naming: this pass does not solve "many series, one line chart"
  any better than colour alone did.
- The stacked-area dashed boundary (visible where a dashed 2nd/3rd series'
  stroke traces the top of its stacked area) reads fine in the screenshots
  taken, but wasn't stress-tested against every possible stacked-area
  colour/opacity combination a consumer might configure via `Colors`/
  `GradientStops`.

## Summary by CodeRabbit

* **New Features**
  * Added reduced-motion support, automatically limiting chart animations when accessibility preferences are enabled.
  * Improved chart theme styling for tooltips, legends, axes, labels, hover states, emphasis, and series.
  * Added decorative patterns for eligible pie, donut, and nightingale charts.
  * Improved line-series differentiation with varied symbols, dash styles, and stroke weights.
  * Refined area-chart gradients, bar styling, corner rounding, and animation consistency across chart types.
  * Chart themes now refresh reliably while preserving custom options and styling.

* **Tests**
  * Expanded coverage for themes, animations, reduced motion, gradients, decorative patterns, and series styling.


Brain2k-0005:
@codex review

PR #396, opened 2026-08-11
feat(gantt)!: promote GanttChart to public API, ship OnTaskUpdate as a real commit gate

Brain2k-0005:
## Summary

Promotes Gantt v3 (working name `Gantt3`) from `[EditorBrowsable(Never)]`/
`PublicAPI.Unshipped` to a real, public, documented component: **`GanttChart`**.
`/components/gantt` now renders it directly.

**This PR contains the one breaking change the ReUI-parity audit identified for
Gantt v3, and it ships in the same release as the promotion on purpose** — see
"The breaking change" below.

## Naming decision

`Gantt3` is a working name; kept it out of the public API. Options considered:
replace the shipped `Gantt` outright, or ship under a clean name with `Gantt`
kept and deprecated in spirit. **Chose the latter: `GanttChart`.**

`Gantt`'s own surface (`JsOnDateChange`/`JsOnProgressChange`/`JsOnTaskClick`,
`OnInitError`, `ZoomLevels`, `DefaultZoom`) has no equivalent on the new engine,
and the two components' defaults differ enough that swapping the export in
place would silently change behaviour under every existing `<Gantt>` usage —
that is a real, consumer-breaking migration, not a rename, and doing it
unilaterally in this task felt like the wrong call. `Gantt` ships completely
untouched, still fully documented at `/components/gantt-legacy`. A full
replacement, if wanted later, should be its own deliberate, versioned migration
with a real deprecation cycle — this PR does not attempt it.

`/components/gantt` itself — every existing link/bookmark — now renders
`GanttChart` directly (`GanttChartPage.razor` declares both
`@page "/components/gantt"` and `@page "/components/gantt-chart"`, since
Blazor allows multiple routes on one component). That's the part that actually
fixes "the owner looked at `/components/gantt`, saw the old component" —
moving `GanttChart` to a brand-new URL and leaving the old renderer on
`/components/gantt` would not have.

## The breaking change

`OnTaskUpdate` converts from a post-commit `EventCallback<GanttTaskUpdate>`
notification into a real pre-commit gate:

```csharp
[Parameter] public Func<GanttTaskUpdate, GanttUpdateResult>? OnTaskUpdate { get; set; }
```

`GanttUpdateResult` (reject / accept / accept-with-adjustment, implicit `bool`
conversion so `OnTaskUpdate="update => true"` still compiles) mirrors
`SchedulerDropResult` — one accept/reject/adjust shape across the library.
`GanttUpdateAdjustment` carries `Start`/`End`/`Progress` (swapping ReUI's
`AllDay` — which Gantt tasks don't have — for `Progress`, the one other field
every gesture through this gate can carry).

**Verified `GanttChart`'s entire surface was still in `PublicAPI.Unshipped.txt`**
before this change (confirmed by grep before touching anything, and by a clean
`-warnaserror` build with the updated `PublicAPI.Shipped/Unshipped.txt` —
RS0016/RS0017 would have fired otherwise). Converting now is free; it becomes
a real breaking change the moment it ships against an already-public
signature, which is exactly why it ships in this same PR.

**Mutation-tested**: forcing the gate to accept unconditionally (bypassing the
`OnTaskUpdate` call entirely) breaks 3 purpose-built tests (reject,
accept-with-adjusted-dates, accept-with-adjusted-progress) plus 4 more across
the existing commit-order/payload suite — 7 failures, each with a concretely
predicted wrong value, all confirmed and reverted.

### What a consumer of the old `Gantt` needs to do

Nothing — `Gantt` is unchanged and still shipped. Consumers who want the new
engine switch `<Gantt ...>` to `<GanttChart ...>` and, if they use
`OnTaskUpdate`, update their handler from `EventCallback<GanttTaskUpdate>` to
`Func<GanttTaskUpdate, GanttUpdateResult>` (a bare `update => true` keeps
existing "just let it through" behavior).

## Docs

`GanttChart`'s demo (`/components/gantt`) covers dependency arrows,
milestones, a live commit-gate demo (rejects past-dated drags, snaps others to
Monday), 7 zoom levels, grouping, progress drag, hierarchy with a
multi-column tree pane and resizable splitter, summary rollups, and the
`GanttSettingsMenu` companion.

Checked against the three defects the owner named on the old page:
- **Bars cut at the container's right edge** — the scroll pane scrolls
  (`overflow:auto`) instead of clipping; verified via Playwright
  (`scrollWidth`/`clientWidth` on the pane, no bar exceeds it unscrolled).
- **Large empty area below the content** — the pane is `max-height`, not a
  fixed `height`; a 6-task demo measured 272px tall against a 380px cap.
- **No task-name pane** — the tree pane renders by default whenever hierarchy
  (`ParentId`) or `GroupBy` is in play; verified present on the rendered page.

## Gates
- Conflict-marker / "adepto" sweep: clean.
- Clean `obj`/`bin` nuke, `dotnet build Lumeo.slnx -warnaserror -m:1`: 0
  warnings, 0 errors (broad `: (error|warning) [A-Za-z]+[0-9]+:` regex count
  also 0).
- Registry regen + `node scripts/algolia/generate-local-index.mjs`: clean,
  new `gantt-chart` entry, all routes verified live.
- `python scripts/docs-parity/check.py`: exit 0 (only a pre-existing,
  unrelated `qr-code` casing warning).
- Full `Lumeo.Tests` suite (net10.0): 7406/7406 passed.
- `Lumeo.RegistryGen.Tests`: 90/90. `Lumeo.Docs.Tests`: 37/37.
- Gantt-scoped E2E suite (isolated `127.0.0.1` ports, `ASPNETCORE_ENVIRONMENT=Development`, docs site + `Lumeo.Tests.ServerHost` both running): 144/144 passed, including the parity/visual suite and the documented `GanttV3StickyHeaderTests` flake (didn't reproduce this run).
- Servers/build-servers shut down by PID (resolved via `netstat`), no stray `obj_*`, no logs committed.

No version bump — release sequencing is the owner's call.

---

## Update: Scheduler self-hosting + expanded Gantt demos

Two follow-up defects found on the deployed preview, fixed on this branch:

### Scheduler renders nothing on the docs site — FullCalendar wasn't self-hosted

`index.html` self-hosted ECharts (`window.lumeoCdn`) for GDPR but never got an
equivalent entry for FullCalendar, so the Scheduler still fetched five modules
from `esm.sh` at runtime — broken on the deployed preview and the same GDPR
gap ECharts self-hosting was meant to close.

`lumeo deps install` doesn't cover this and can't as-is: it explicitly detects
esm.sh bare-specifier URLs (`https://esm.sh/@fullcalendar/core@6`) as
"esm-only — cannot self-host" and skips them (no downloadable single file to
grab).

Fixed by vendoring the five FullCalendar v6.1.21 packages plus their `preact`
10.12.1 dependency as their **unmodified npm-published ESM entry points**
under `wwwroot/lib/lumeo-vendor/` — not an esm.sh bundle re-export, which
would have needed either its own second `preact` copy per package (multiple
live Preact instances, a real correctness risk) or duplicated core internals.
Each package's own internal bare-specifier imports (`preact`,
`@fullcalendar/core/*`, `@fullcalendar/daygrid/internal.js`) are rewritten to
the local vendored paths at vendor time — what a bundler would otherwise do.
An import map wasn't an option: the docs WASM build owns the page's sole
`<script type="importmap">` and overwrites it at publish. Locale packs needed
no separate wiring — `scheduler.js` already resolves them relative to
wherever `fullCalendarCore` itself resolved to (fixed in an earlier PR).

Verified with a real Chromium (Playwright): **zero requests to `esm.sh`** on
`/components/scheduler`, and a freshly-initialized `locale: 'de'` /
`locale: 'fr'` instance renders correctly localized day headers ("Mo Di Mi…" /
"lun. mar. mer…") — confirms self-hosting core didn't silently fall back to
English.

### Gantt demo page — expanded, and a real defect found + fixed

The promoted `GanttChart` page showed too little of what three phases of work
built. Enriched the flagship demo into a realistic ~40-day, 4-phase project
plan (hierarchy + cross-phase dependencies + milestones) and added four new
demos: **today marker + off-day shading** (HalfDay zoom — `NowIndicator` only
renders at sub-day granularity), **leaf-row checkbox selection**, **tree-row
drag reorder**, and a custom **`RowTemplate`**. ISO week bands are already
visible automatically in every `Day`-mode demo's upper header (Phase 3); the
floating zoom control, `ColorByGroup`, and infinite scroll were already
demonstrated and are called out more explicitly in captions now.

While verifying against the "no reintroduced defects" bar, found a real,
**pre-existing** bug affecting every Gantt demo on this page (not introduced
by the new ones, and not caught by the original PR's own defect check —
apparently narrower/wider datasets hit it differently): `ComponentDemo`'s
preview box centers its content with `flex items-center justify-center`
without stretching children to full width. `GanttChart`'s own internal
scroll-host sizing depends on that width cascading down correctly; without it,
the auto-center-on-today scroll landed on an element that was never actually
constrained, so on narrower-range demos the visible viewport showed a blank
grid at the *wrong month* with the tree pane scrolled fully off-screen — the
exact "large empty area" / "content not visible" failure class this PR's own
gate was meant to catch. Fixed with `Class="w-full"` on each demo's chart (or
its wrapping `Stack`), which forces the width cascade through the flex box —
matching the layout already working correctly on the dedicated E2E preview
page. Also fixed two `&amp;` literals in `Title="..."` attributes (a plain
string parameter, not HTML) that were rendering as the literal text "&amp;"
on the page.

Verified via Playwright against all 14 demo instances: every chart shows
today's column, at least one task bar, and (where applicable) a correctly
positioned tree pane on initial load; zero console/page errors;
`#blazor-error-ui` never visible; the page itself never grows wider than the
viewport (no accidental page-level horizontal scroll).

### New E2E coverage

`Lumeo.Tests.E2E/Smokes/GanttChartPageDemosTests.cs` — asserts all 14 demo
sections render with a live chart (per-section, not just "the page
responded" — a Blazor render exception kills everything below it), plus
targeted checks for the row-reorder tree pane, the row-selection counter, the
settings popover, and the now-line/off-day markers. Timeout budgeted from a
measured 16.3–16.9s local full-page render (14 live `GanttChart` instances,
vs. the existing suite's heaviest page at 5 instances / 30000ms) — 45000ms.
5/5 passing against an isolated dev-server port.

### Gates (this update)
- Conflict-marker / "adepto" sweep: clean.
- Clean `obj`/`bin` nuke, `dotnet build Lumeo.slnx -warnaserror -m:1`: 0
  warnings, 0 errors (broad regex cross-check also 0).
- Registry regen + `node scripts/algolia/generate-local-index.mjs`: clean,
  no diff — vendored FullCalendar/preact JS does not leak into the registry
  or the search index.
- `python scripts/docs-parity/check.py`: exit 0 (same pre-existing, unrelated
  `qr-code` casing warning as before).
- `Lumeo.Tests` (net10.0): 7406/7406 passed. `Lumeo.Docs.Tests`: 37/37 passed.
- Docs-site E2E `Smokes/` suite (isolated `127.0.0.1` port,
  `ASPNETCORE_ENVIRONMENT=Development`, excluding two pre-existing specs that
  target a separate `Lumeo.Tests.ServerHost` process not started for this
  run): 39/39 passed, including the new Gantt demos suite.
- Build servers shut down by PID (resolved via `netstat`), dev server on an
  isolated port stopped, port 5287 never touched, no stray `obj_*`, no logs
  committed. Vendored FullCalendar/preact JS is a deliberate commit.

No version bump.

### Wheel zoom + arrow-key navigation

Two ReUI-parity interactions the v3 engine had **zero** code for (no wheel
handling anywhere in `gantt-v3.js`; keyboard support was Enter/Space
activation only, gated on `OnTaskClick`).

**Wheel zoom** — Ctrl/Cmd + wheel over the timeline zooms, anchored on the
pointer position; a bare wheel keeps scrolling the page/pane exactly as if
no Gantt were on it (the modifier check is the first thing the listener
does, unconditionally, every event). **Stepped, not continuous**: `GanttChart`'s
zoom model is already a discrete Day/Week/Month/[Quarter]/Year ladder, unlike
ReUI's continuous 0.5–3 factor grafted onto a single view — a continuous
factor would be an orthogonal, much larger change (a new scale dimension
every pixel computation would need to multiply through) for a gesture that,
stepped, already reads as "zoom". Steps through the exact same resolved
zoom-level list the toolbar/floating `GanttZoomControl` already step
through — never a second, independently-tunable list. At the resolved
list's own zoom limits the gesture is left alone entirely (no
`preventDefault`), so the browser's own Ctrl/Cmd+wheel page-zoom takes over
instead of the gesture being silently swallowed — decided synchronously in
JS (from an options bag C# pushes on every `ViewMode`/`ZoomLevels` change),
since an async decision can't un-ring a wheel event the browser already
processed.

Anchoring reuses the *existing* view-mode capture/recenter pipeline
(`GanttChart.ReconcileAsync`/`ApplyViewModeIntentAsync`) rather than
duplicating it: two new optional parameters
(`explicitCapturedCenter`/`explicitOffsetOverride`, both `null` for every
pre-existing caller — toolbar, imperative, parameter, task-change triggers
are all byte-identical) let wheel-zoom skip the live-viewport-center capture
(it already knows its own pointer-derived anchor date) and recenter through
a **new** `GanttV3ScrollToOffsetAsync` interop call (places a date at the
pointer's own local viewport offset) instead of the existing
`GanttV3ScrollToXAsync`/`centerOn` (viewport-*center* placement) every other
`ViewMode` trigger keeps using, unchanged.

**Arrow-key navigation** — a roving tabindex over bars
(`GanttBar.FocusedTaskId`, wired unconditionally by `GanttTimeline`, not a
consumer opt-in): plain arrows move focus to the previous/next visible row
(Left/Right alias Up/Down — one task per row, no distinct "sideways" bar to
move to); `Shift`+arrow nudges the *focused* task's own schedule by a day
(Left/Right move both edges, Up/Down grow/shrink the End edge — keyboard
resize covers the End edge only, mirroring v2's own end-only resize
precedent) through the **same `OnTaskUpdate` commit gate a mouse drag
uses**, tagged the new `GanttTaskUpdateSource.Keyboard` (ReUI tags a
keyboard-originated edit with one flat "keyboard" source rather than
separate per-gesture keyboard members; this does the same). An aria-live
region announces focus-boundary/schedule-nudge outcomes — the Scheduler's
first-party view engine (`feat/scheduler-first-party`) isn't merged into
this branch and has no announcement precedent of its own yet, so this
follows the repo's actual existing convention instead
(`Lumeo.Sortable`/`Lumeo.DataGrid`'s sr-only `aria-live="polite"` span).

Building this surfaced a real, independent bug: since `GanttTimeline` now
*always* wires `OnKeyNavigation`, removing a consumer's `OnTaskClick`
mid-flight no longer means "nothing wants the prevent-default-keys
registration anymore" (arrow nav still does) — the desired *rule
combination* can drift while a register call is in flight without ever
reaching the "neither wants it" branch a single-bool design could get away
with. Fixed in `GanttBar.OnAfterRenderAsync`: re-registers with the live
combination when it's drifted from what was actually sent, instead of
leaving a stale rule set in place (which would have kept Space suppressed
after `OnTaskClick` was gone). The pre-existing round-17 regression test for
this class of bug now renders `GanttBar` directly (isolating the scenario it
actually covers — `OnTaskClick` as the *sole* driver) with a new sibling
test covering the drift case.

`activeDragGestureCount` (issue #400, page-global gesture-suppression
counter): neither feature touches it — wheel zoom and keyboard nudges are
single-shot commits, not pointer gestures, so there's no ghost/pointer
capture/cleanup path to interact with the counter at all. Not modified;
scoping it per-instance is a separate, larger refactor out of this task's
two-feature scope.

### Gates (this update)
- Conflict-marker / "adepto" sweep: clean.
- Clean `obj`/`bin` nuke, `dotnet build Lumeo.slnx -warnaserror -m:1`: 0
  warnings, 0 errors.
- Registry regen (`dotnet run --project tools/Lumeo.RegistryGen`) +
  `node scripts/algolia/generate-local-index.mjs`: clean; picked up the new
  `GanttChart.WheelZoom` param and the new demo section.
- `python scripts/docs-parity/check.py`: exit 0 (same pre-existing,
  unrelated `qr-code` casing warning).
- Full filtered suite (`dotnet test Lumeo.slnx -c Release --no-build -f
  net10.0 --filter "FullyQualifiedName!~Lumeo.Tests.E2E&Category!=EjectGateFull"`):
  **7439/7439 passed** (+33 tests vs. this branch's own pre-existing
  baseline, which had exactly one failure — a real regression from the
  `OnKeyNavigation`-always-wired change above, root-caused and fixed, not
  papered over; see the drift-registration fix). Mutation-tested the
  keyboard commit gate directly: forcing `GanttChart`'s own verdict
  evaluation to bypass a `Reject` and commit anyway makes the corresponding
  bUnit test fail as expected (predicted-vs-actual dates diverge); reverted
  immediately after confirming.
- **Local E2E**: hit the same wall a predecessor already flagged — this
  sandbox's `Lumeo.Tests.ServerHost` fixture host (real Blazor Server +
  SignalR circuit) never establishes its WebSocket connection under
  Playwright here, so `[data-testid='gantt-v3-root'] div[style*='overflow']`
  never appears and every spec against it times out. Confirmed this is a
  sandbox limitation, not a regression from this change: an **unmodified**
  pre-existing spec (`GanttV3ClickActivationTests`, untouched by this PR)
  fails identically against the same isolated host. New specs were written
  end-to-end anyway (`GanttV3WheelZoomTests`/`GanttV3KeyboardNavigationTests`
  under `Lumeo.Tests.E2E/Gantt/`, plus a new `data-testid="gantt-v3-viewmode"`
  sink on the E2E fixture page) — they build clean and are ready to run
  wherever the SignalR wall doesn't apply; reporting this honestly rather
  than fabricating a pass.
- Build servers shut down, isolated ServerHost (port 5301, never 5299/5287)
  stopped by its own PID via `netstat`, no stray `obj_*`, no logs committed.

No version bump.


Brain2k-0005:
@codex review

PR #397, opened 2026-08-11
fix(docs): stop three visible defects on the docs site

Brain2k-0005:
## Summary

Three independent, real bugs the repo owner could see live on the docs site today. All three are unrelated to the ongoing chart bar-geometry, Gantt v3, and Scheduler-preview work landing on other branches.

- **Demo container (`Shared/ComponentDemo.razor`, used by every `/components/*` preview)**: paired an unbounded `min-h` box with `overflow-y-auto`. Harmless for tall *visible* content, but on chart pages it clipped content that was never visible in the first place — `Chart.razor`'s screen-reader-only data table is an absolutely-positioned `<table>`, and the CSS table-sizing algorithm treats `sr-only`'s `width/height:1px` as a *minimum*, not an exact size, so the invisible table still inflated `scrollHeight` and popped an unwanted inner scrollbar. `overflow-y-auto` → `overflow-y-hidden`; `overflow-x-auto` untouched (still legitimate for wide tables).
- **Gantt** (the old `<Gantt>`/`gantt-v2.js` component — still the default at `/components/gantt` on `master`): the "Basic project plan" / "With progress bars" demos' data starts exactly at "today", so the center-on-today initial scroll wasted half the viewport on empty padding and cut the "Design" bar off mid-label. Fixed with a narrower `ColumnWidth` on just those two demos — deliberately not touching `gantt-v2.js`'s shared scroll/padding behavior itself (real drag-to-reschedule consumers rely on it, and it's pinned by the v2/v3 parity visual baselines).
- **Scheduler** (FullCalendar-backed, default at `/components/scheduler`): month-view event titles truncated mid-word with no ellipsis; fixed with `text-overflow:ellipsis` + a native `title` tooltip carrying the full text. Separately, event text color was hardcoded to `--color-primary-foreground` regardless of a per-event `Color` override, so a background whose real contrasting foreground differs could render illegibly (measured live, dark mode: `Offsite`'s `--color-accent` background against the forced text ≈ 1.19:1 contrast). Fixed generically — `scheduler.js` now tags `Color`-overridden events with a class that pairs them with their *own* `-foreground` token. The docs fixture also moved off `--color-destructive` (danger red for a routine meeting) and `--color-accent` (a hover-surface grey indistinguishable from the page background in dark mode) onto `--color-info` / `--color-success`.

## Test plan

- [x] Added `tests/Lumeo.Tests.E2E/Smokes/ComponentDemoOverflowTests.cs`, `GanttDemoBarOverflowTests.cs`, `SchedulerEventColorAndTruncationTests.cs` — each asserts real rendered geometry/color (computed `overflow-y`, bar bounding-box vs viewport, WCAG contrast ratio, `scrollWidth` vs `clientWidth`), not class strings.
- [x] Disable-check on each fix reproduces the exact predicted broken value (`overflow-y: auto`; ~260px bar overflow; ~1.19:1 contrast / `text-overflow: clip`) before failing; passes after restoring the fix.
- [x] `dotnet build Lumeo.slnx -warnaserror -m:1` — 0 errors, 0 warnings (clean rebuild, broad regex cross-check).
- [x] Registry regen (`dotnet run --project tools/Lumeo.RegistryGen`) + docs registry copy + Algolia local index — picked up the new E2E test coverage and updated demo code snippets only.
- [x] `python scripts/docs-parity/check.py` — exit 0.
- [x] Filtered suite (`dotnet test Lumeo.slnx -c Release --filter "FullyQualifiedName!~Lumeo.Tests.E2E&Category!=EjectGateFull"`) — 7393/7393 passed.
- [x] Full local E2E suite on isolated ports — 219/222 passed. The 3 failures: `HomePageVisualTests` (documented Windows-local pixel-diff noise, green in CI) and 2 Scheduler smokes that time out waiting on FullCalendar's `esm.sh` CDN import chain under this run's heavy concurrent load — confirmed via direct `curl`/isolated browser probes that this reproduces identically on unmodified `master`, i.e. it's environment/session flakiness, not a regression from this change.

## Summary by CodeRabbit

- **Enhancements**
  - Scheduler events now use clearer theme-based colors with improved text contrast.
  - Long event titles in narrow calendar cells are truncated with ellipses and provide full-text tooltips.
  - Event tooltips display the scheduled time and title.
  - Documentation previews support horizontal scrolling without unnecessary vertical scrollbars.
  - Gantt demo bars remain visible within the initial viewport, with improved day-view sizing.
- **Tests**
  - Added end-to-end coverage for scheduler accessibility, truncation, overflow behavior, and Gantt visibility.


Brain2k-0005:
@codex review

PR #398, opened 2026-08-11
feat(theme): give the chart palette real hue separation and chroma

Brain2k-0005:
## Summary

The chart palette had two measured problems:

- `--color-chart-1` resolved to the exact same oklch value as `--color-primary`
  in the default light theme (chroma 0.0059, effectively achromatic), so a
  single-series chart — and the Gantt's `ColorByGroup` fallback, which reads
  `--color-chart-1..5` — both rendered grey.
- `--color-chart-4`/`--color-chart-5` (amber vs orange) sat only ~29deg apart in
  hue and read as indistinguishable, and were byte-identical between light and
  dark mode.

This is a **deliberate divergence from shadcn's chart defaults**, approved by
the owner ("bei der Palette kann ruhig bunt sein" — color can be vivid). All
other primitives (`--primary`, `--secondary`, `--accent`, `--destructive`,
`--success`, `--info`, `--warning`, `--muted`, `--border`, `--ring`, etc.) are
**untouched** — this PR only changes the five `--color-chart-N` tokens, in
`lumeo.css` and the 8 named theme files. The `var(--chart-N, <fallback>)`
indirection is preserved everywhere, so consumers can still override every
token.

### Palette design

- **Default / Zinc** (both have an intentionally achromatic `--primary`):
  `chart-1` no longer mirrors `--primary`. All 5 tokens now use an
  independent, evenly-spaced (72deg apart) vivid hue wheel — blue, teal,
  green, amber, magenta — tuned separately for light and dark.
- **The 7 chromatic named themes** (amber, blue, green, orange, rose, teal,
  violet) originally kept `chart-1 = <theme's own --primary>` here, so a
  theme's charts read as belonging to that theme — see the "Update" section
  below for why that was reverted and replaced with a fully independent
  palette per theme.

### Guard test

Adds `ChartTokenGuardTests` (scans raw file text, same approach as
`RadiusTokenGuardTests`) pinning the palette's *properties*, not its exact
values, across all 9 files (base + 8 themes) x 2 modes:
- every block's 5 chart hues stay >=40deg apart pairwise
- every chart token clears a chroma floor (bans near-achromatic greys)
- light and dark differ per token
- `--color-chart-1` no longer equals `--color-primary` in the base light theme

Verified the guard catches the regression: reverting `chart-1` to its old
`oklch(0.2103 0.0059 285.88)` value fails both the chroma-floor check and the
chart1-vs-primary check with the exact predicted values, before being
restored.

## Test plan

- [x] Conflict-marker + literal sweep — clean
- [x] `dotnet build Lumeo.slnx -warnaserror -m:1` — 0 warnings, 0 errors
- [x] Registry regen + copy + `generate-local-index.mjs` — only timestamps and
      the new test file's cross-reference changed
- [x] `python scripts/docs-parity/check.py` — exit 0
- [x] Full filtered suite (non-E2E): 7567/7567 passed
- [x] Local E2E (isolated ports 5290/5299, `ASPNETCORE_ENVIRONMENT=Development`):
      230/232 passed. The 2 failures are pre-existing and unrelated:
      `HomePageVisualTests` (known ~8% diff on Windows locally, green in CI —
      not recaptured per standing guidance) and a flaky pointer-drag splitter
      test that passes in isolation. `GanttParityVisualTests` (11 baselines)
      all passed as-is at the time — **superseded by the "Update" section
      below**, where the Gantt's default bar fill was decoupled from
      `--primary` too and 8 of those baselines were legitimately recaptured.
- [x] Visually verified `/components/charts/{bar,line,pie}` and
      `/components/gantt` in light + dark, plus the `blue` and `rose` named
      themes, via a headless Chromium/Playwright script (own docs server on
      an isolated port; owner's port 5287 untouched)

---

## Update: chart colours are now independent of `--primary` everywhere

The initial palette work above fixed the **base/zinc** themes but left two
couplings in place, both now fixed:

- The **7 named themes** (amber/blue/green/orange/rose/teal/violet) set
  `chart-1 = --primary` literally and rotated `chart-2..5` around it — so
  switching a theme's accent silently repainted every chart's first data
  series. Each theme now gets an independently-derived chart palette (the
  same blue/cyan/green/amber/magenta hue family base/zinc already use),
  rotated so `chart-1`'s hue sits well clear of that theme's own primary hue,
  in both light and dark.
- The **Gantt's default bar fill** (no `BarColor`, no `ColorByGroup`) fell
  back to `var(--color-primary)`, rendering grey/near-black bars in the base
  theme. It now falls back to `var(--color-chart-1)`, so a plain Gantt
  inherits chart theming like every other chart component; `ColorByGroup`
  is unaffected.

Also fixed while sweeping for the same pattern: `WaterfallChart.IncreaseColor`
defaulted to `var(--color-primary)` (now `var(--color-chart-1)`), and
`echarts-interop.js`'s `chart1` CSS-var resolution had `--color-primary` as a
middle fallback before its hex literal (removed).

`ChartTokenGuardTests` gained an independence assertion across all 9 palette
files x both modes: no chart token may equal that file's `--color-primary`.

Recaptured 8 `GanttParityVisualTests` baselines whose default bars actually
changed pixels (`gantt-v3-{Day,Week,Month}` plus 5 T10 v3-only baselines with
visible default-colored bars). Left v2, `v3-Year`, `v3-t10-quarter-mode`,
`v3-t10-offscreen-chip` and `v3-t7-chrome` (`ColorByGroup`) untouched —
confirmed against the old baselines that those don't exercise the changed
path or don't show enough bar area to cross the diff threshold.

Verified by eye with a headless Chromium/Playwright script against a docs
server on an isolated port (owner's 5287 untouched): `/components/charts/bar`,
`/components/charts/pie`, and the Gantt v3 preview, in light/dark and the
blue/green/rose named themes — default Gantt bars and chart series colours no
longer track the active theme's brand/accent colour.

## Summary by CodeRabbit

* **Style**
  * Updated chart color palettes across light and dark themes for improved visual distinction and consistency.
  * Refined chart colors across all supported theme variants.

* **New Features**
  * Waterfall charts and Gantt bars now use the first chart-palette color by default.

* **Tests**
  * Added automated checks for five distinct, vivid chart colors in both light and dark modes.

* **Chores**
  * Refreshed generated registry metadata and timestamps.


Brain2k-0005:
@codex review

PR #401, opened 2026-08-12
Complete AlertDialog/Drawer size lever; fix Tabs density height bug

Brain2k-0005:
## Summary
- **AlertDialog** and **Drawer** had no size control at all. Both are structural twins of components that already have one (`Dialog`/`DialogSize`, `Sheet`/`SheetSize`) — this is family completion, not a `Lumeo.Size` migration, since `Full` is a layout intent rather than a scale value (the same documented reason `DialogSize`/`SheetSize` stay outside `Lumeo.Size`).
  - `AlertDialogContent.Size` (`AlertDialogSize`: `Sm`/`Default`/`Lg`/`Xl`/`Full`) mirrors `DialogContent.Size` exactly.
  - `DrawerContent.Size` (`DrawerSize`: `Sm`/`Default`/`Lg`/`Xl`/`Full`) mirrors `SheetContent.Size`; on Left/Right it only lifts the max-width cap (base `w-3/4` unaffected, mirroring `SheetSize.Full`'s own Left/Right precedent); on Top/Bottom, `Full` additionally drops the permanent `mt-24` "peek" offset for a true mobile-fullscreen takeover.
  - Both default to the pre-existing value, reproducing today's rendering byte-for-byte for a consumer who sets nothing.
  - Every Tailwind class token used already exists in both shipped CSS bundles (verified with a literal Python substring scan) — no CSS rebuild needed.
- **Tabs density bug**: `TabsList`'s track height was hardcoded (`h-9`/`h-10`) and never followed the ambient `Density`, even though `TabsTrigger`'s own padding already does — at Compact/Spacious the triggers changed size inside a track that stayed fixed. `TrackHeightClass` derives from the same `(Variant, Orientation, Density)` inputs `PadClass` already resolves.

## Test plan
- [x] New `AlertDialogSizeTests` (8 tests) — exact-token assertions per size, default-unchanged proof, disable-check verified (5/8 fail against the pre-fix single hardcoded class).
- [x] New `DrawerSizeTests` (18 tests) — per-Side/Size exact-token coverage, `mt-24` behavior, disable-check verified (11/18 fail against the pre-fix hardcoded classes).
- [x] New `TabsListDensityHeightTests` (22 tests) — 12-combination (Variant × Density) matrix, vertical-untouched proof, 2 predicted-vs-actual disable-check cases (both verified failing against the pre-fix hardcoded heights).
- [x] Existing 24-combination `TabsTriggerPaddingTests` matrix untouched and still green.
- [x] `dotnet build Lumeo.slnx -warnaserror -m:1` — 0 warnings, 0 errors.
- [x] Registry regenerated + copied + local search index rebuilt; `docs-parity/check.py` — OK.
- [x] Full filtered suite (`Lumeo.Tests`, excluding E2E): 7605/7605 passed.
- [x] Local E2E (isolated port 5295, excluding `Gantt/*`): 92/93 passed — the one failure is the known `HomePageVisualTests` Windows-local pixel-diff noise called out in the task (not recaptured).
- `tests/Lumeo.Tests.E2E/Gantt/*` was not run — its fixture host needs SignalR WebSockets, which are blocked in this environment (consistent with prior reports).

## Summary by CodeRabbit

* **New Features**
  * Added configurable sizes for alert dialogs and drawers, including small, large, extra-large, and full-screen options.
  * Improved tab sizing so horizontal tab tracks adapt to the selected density.
  * Added documentation examples showcasing all dialog and drawer size variants.

* **Bug Fixes**
  * Ensured full-size panels use appropriate viewport dimensions and spacing.

* **Tests**
  * Added coverage for dialog sizing, drawer sizing, and density-aware tab heights.


Brain2k-0005:
@codex review

PR #402, opened 2026-08-12
Migrate Badge/BorderBeam/Progress size scales onto Lumeo.Size

Brain2k-0005:
## Summary

Three components carried their own size type with no non-scale value to justify it -- oversights the 3.0 `Lumeo.Size` unification never swept:

- **`Badge`** -- `BadgeSize { Sm, Md, Lg }`, a pure subset of `Lumeo.Size`'s own names.
- **`BorderBeam`** (Lumeo.Motion) -- `BeamSize { Sm, Md, Lg }`, same shape.
- **`Progress`** -- an **untyped `string Size`** taking `"Small"`/`"Default"`/`"Large"`. No enum, no compile-time safety -- a typo silently fails at runtime and the preview falls back to `Md`.

`Size.cs` documents exactly two justified exceptions to the unified scale (`Button.ButtonSize` for its `Icon` shape value, `Dialog`/`Sheet` for their `Full` layout-intent value). These three had no such justification, so they're migrated onto `Lumeo.Size` here, all 7 rungs (Xxs-Xxl).

**Why now, not later:** every day this ships as-is, more consumers write code against the current (wrong) shape. Doing the break early, before the surface area grows, is cheaper than doing it after a wider release.

## This is a public API change

| Component | Before | After |
|---|---|---|
| `Badge.Size` | `Badge.BadgeSize` (`Sm`/`Md`/`Lg`) | `Lumeo.Size` (`Xxs`-`Xxl`) |
| `BorderBeam.Size` | `BorderBeam.BeamSize` (`Sm`/`Md`/`Lg`) | `Lumeo.Size` (`Xxs`-`Xxl`) |
| `Progress.Size` | `string` (`"Small"`/`"Default"`/`"Large"`) | `Lumeo.Size` (`Xxs`-`Xxl`) |

**What a consumer must do:** replace `Badge.BadgeSize.Sm/Md/Lg`, `BorderBeam.BeamSize.Sm/Md/Lg`, and `Size="Small"/"Default"/"Large"` with `Lumeo.Size.Sm/Md/Lg`. Rendering at those three rungs is unchanged (verified byte-for-byte, see Test plan). No other parameters changed.

**On `Progress`'s string parameter specifically:** this is a hard break rather than keeping the string as an `[Obsolete]` alias alongside the new typed parameter. The whole point of migrating it is removing the class of defect a permanently-dangerous, typo-prone string parameter enables -- keeping it around (even deprecated) preserves that footgun. None of the other 17 already-migrated components in this campaign carried a compat shim either, so this keeps the pattern consistent.

No version bump in this PR, per project convention -- tagging/release is a separate, owner-confirmed step.

## Test plan

- [x] `Badge`/`BorderBeam`/`Progress`: `Sm`/`Md`/`Lg` (and old `"Small"`/`"Default"`/`"Large"`) pinned with exact space-delimited class-token assertions (never `Assert.Contains`) proving byte-identical output to the pre-migration values.
- [x] Disable-check per component: temporarily removed one rung's mapping, confirmed the concretely predicted wrong fallback value, restored.
  - Badge `Lg` padding: predicted fallthrough to Md's `px-2.5 py-0.5` -- confirmed, restored.
  - BorderBeam `Sm`: predicted fallthrough to Md's `1.5px` -- confirmed, restored.
  - Progress circular `Sm`: predicted fallthrough to Md's `width: 56px; height: 56px` -- confirmed, restored.
- [x] `SizeScaleOrderingTests` extended for all three: monotonic Xxs<Xs<Sm<Md<Lg<Xl<Xxl ramp, with documented ties at the low end (Xxs/Xs, and Badge's Xxs/Xs/Sm at py-0).
- [x] `Badge`/`BorderBeam`/`Progress` docs pages extended with a full 7-rung interactive `Segmented` picker using real `Lumeo.Size` member names (not the `"Small"`/`"Medium"`/`"Large"` `Enum.TryParse` bug pattern found and fixed on four other pages in an earlier campaign).
- [x] `PublicAPI.Unshipped.txt` updated (`*REMOVED*` + new shape) for `Lumeo` and `Lumeo.Motion`; no version bump.
- [x] Full clean solution build, `-warnaserror -m:1`: 0 errors, 0 warnings.
- [x] `scripts/docs-parity/check.py`: OK, in parity.
- [x] Full filtered suite (excl. E2E + EjectGateFull): 7595/7595 passed (delta: +7 vs. pre-change baseline, all new Badge preservation tests).
- [x] Local E2E on an isolated port (5390, `5287` untouched), excluding the known-broken `Gantt/*` suite (SignalR WebSockets unavailable in this environment) and known-flaky `HomePageVisualTests`/`GanttV3StickyHeaderTests`: 92/92 passed, incl. `CatalogPageRendersTests` sweeping all doc pages.


Brain2k-0005:
@codex review

Issue #390, opened 2026-08-10
Gantt v3: mount-time scroll-to-today can fire twice and clobber a later scroll write


Brain2k-0005:
The Gantt's mount-time "scroll to today" can fire **twice** on a Blazor Server circuit,
and the second write can silently clobber a scroll position set by the consumer (or by a
test) immediately after mount.

## Cause

`GanttTimeline.OnAfterRenderAsync` sets the sentinels that mean "the initial scroll is
done" (`IsScrollHostReady`, `_lastSeenScrollHostId`, and the related latch) **only after
several awaited interop round-trips** — `SyncDragRegistrationAsync`,
`ReconcileHeaderScrollSyncAsync` and others. Over SignalR those awaits are not instant, so
a second render can dispatch while the first is still mid-await, read the still-stale
sentinels, and independently conclude that an initial scroll is owed.

Both calls target the same pixel value, so the duplicate is invisible in the common case.
It becomes visible when something else writes `scrollLeft` right after mount: the second
call is scheduled via `requestAnimationFrame` and lands a few milliseconds later,
overwriting that write.

The existing latch proves the *first* call landed. It does not prove that no further call
is still in flight.

## How it surfaced

`GanttV3NavTests.Today_button_scrolls_the_today_marker_into_view` sets `scrollLeft = 0` as
its precondition, then asserts the today marker is out of view before clicking Today. At a
~15–20% rate the duplicate scroll restored the centred position first, so the precondition
failed with `viewport ratio 1` and the test could no longer exercise what it exists to
verify.

Diagnosed by injecting a temporary `scrollLeft` write-logger plus server-side tracing into
`OnAfterRenderAsync`, which showed two independent `ScrollToCenterAsync` calls per mount.

## Current state

PR #387 works around this **in the test only** — the `scrollLeft = 0` override retries up
to three times, absorbing one extra in-flight write. Assertion and precondition semantics
are unchanged, so no coverage was retired. The underlying race is untouched and
pre-existing.

## Suggested fix

Claim the "initial scroll owed" state **before** the first await rather than after the
round-trips complete, so a concurrently dispatched render sees it already taken. This is
the same shape as the range-extension in-flight guard fixed during the Phase 3 review
(`_rangeExtensionInFlight` was set after its first await for the same reason).

Worth checking whether other `OnAfterRenderAsync` paths in the component share the pattern.

## Impact

Low severity, non-zero: an extra interop round-trip on every mount, and a consumer that
[...]

PR #405, opened 2026-08-12
fix(gantt-v3): close the residual mount-scroll race that silently drops infinite-scroll extensions

Brain2k-0005:
## Master's gate is red

PR #396's merge (`bc279e79`) broke the E2E gate: `GanttV3InfiniteScrollTests`'
leading-edge extension specs fail with `scrollWidth never grew ... the extension
never committed within the retry budget`. Every push/PR CI run since has
inherited the failure (four open PRs are blocked behind it).

## Root cause

`#390` fixed `GanttTimeline.OnAfterRenderAsync`'s mount-time double-scroll by
claiming `_lastConsumedScrollRequestId` **before** the first `await` instead of
after, closing the race where a second, `GanttChart`-forced render (dispatched
once `ScrollHost` captures) re-derives a stale "is a scroll still owed" answer
and independently re-scrolls.

It left a sibling race open: `firstRender`'s own `initialIntentPending`
decision is captured **once**, up front, and acted on unconditionally once its
several interop round-trips resolve (`SyncDragRegistrationAsync`,
`SyncBarContextMenuRegistrationAsync`, `SyncWheelZoomRegistrationAsync`,
`ReconcileHeaderScrollSyncAsync`) — even when the second, faster-racing render
already issued the equivalent centering scroll first.

Confirmed live with diagnostic logging (stack trace on the arming call): both
passes call `ScrollToCenterAsync` for the *same* request id, back to back,
each re-arming `_suppressNextExtensionReport`. Harmless on a fast local run
(the redundant call resolves before any later report arrives), but under CI's
slower interop timing — lengthened further by this same PR's own added
`SyncWheelZoomRegistrationAsync` round-trip — the redundant call can still be
in flight when a genuine near-leading-edge scroll report arrives, gets wrongly
suppressed, and (since nothing else re-triggers it) never retries.

This also explains why PR #396's own CI run was green on its merge ref while
the squashed push to master was red: it's a genuine timing-dependent race, not
a deterministic bug — the same code can pass or fail across separate CI
executions depending on fine-grained scheduling. A green PR run is not a
guarantee master stays green.

## Fix

A claim ticket (`_mountScrollClaim`) scoped to *only* that one mount-time
race: `firstRender`'s own trailing scroll only fires if nothing else has
already claimed the same request id first. The non-firstRender branch
(`Today`-changed / `scrollHostChanged`) keeps scrolling unconditionally
exactly as before.

An earlier attempt gated on "has this request id ever been issued" instead —
that broke `GanttV3CodexRound2Tests` / `GanttV3CodexRound14Tests`' coverage of
a legitimate *later* re-scroll that reuses the same id (`Today` changing after
mount never bumps `ScrollToTodayRequestId`). Caught locally before this PR,
fixed by narrowing the guard to the specific two call sites that can actually
race for the same id.

## Verification

- Predicted-vs-actual disable-check: an artificial delay ahead of
  `firstRender`'s trailing action reproduces `scrollWidth never grew ...`
  on the pre-fix code, and passes cleanly once the fix is applied.
- `GanttV3InfiniteScrollTests`: 5/5 repeated local runs, clean.
- Full `Lumeo.Tests` (non-E2E): 7653/7653.
- Full `Gantt`/`Smokes`/`Scheduler` E2E (`CI=true`, matching the real gate):
  249/250 — the one failure is a pre-existing wheel-zoom flake, reproduced
  identically on unmodified master.
- Build: 0 warnings, 0 errors. Registry/docs-parity: no drift.

Does not revert #390 (the mount-scroll race it fixed stays fixed) and does not
widen any timeout.

Generated with Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Gantt timeline positioning during initial loading and asynchronous updates.
  * Prevented duplicate automatic centering that could cause unnecessary scrolling or visual movement.
  * Preserved expected scrolling behavior for subsequent timeline interactions and updates.

### DIFF
--- a/src/Lumeo.Gantt/UI/GanttChart/GanttTimeline.razor
+++ b/src/Lumeo.Gantt/UI/GanttChart/GanttTimeline.razor
@@ -1120,6 +1120,75 @@
     // more range."
     private bool _suppressNextExtensionReport;
 
+    // Bug fix (residual #390 double-scroll — reconciling the mount-scroll
+    // dedup with infinite scroll's own suppression flag above): #390 closed
+    // the race where a SECOND render, dispatched while firstRender's own
+    // pass is still mid-flight in its several awaited interop round-trips,
+    // independently re-derives "is a scroll still owed" from the (at that
+    // point not-yet-claimed) _lastConsumedScrollRequestId and wrongly
+    // concludes one is. It did NOT close a sibling race: firstRender's own
+    // "initialIntentPending" decision is captured ONCE, in a local variable,
+    // before any of those awaits, and is acted on UNCONDITIONALLY once they
+    // all resolve — even if a DIFFERENT, faster-racing render already issued
+    // the equivalent scroll in the meantime. That other render is real:
+    // GanttChart's own firstRender OnAfterRenderAsync calls StateHasChanged()
+    // BEFORE its own awaits specifically so GanttTimeline gets a "second
+    // chance" once ScrollHost captures (see this component's own
+    // OnAfterRenderAsync remarks) — that second, non-firstRender render
+    // reaches its scrollHostChanged branch, which (correctly, by design —
+    // see its own remarks) issues the centering scroll WITHOUT gating on
+    // intentPending, since a host becoming ready is itself a re-scroll
+    // trigger independent of the bookkeeping flag. Confirmed live (this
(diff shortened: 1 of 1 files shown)


Brain2k-0005:
@codex review

PR #406, opened 2026-08-12
feat(gantt): add the eight ReUI-parity data-* styling hooks

Brain2k-0005:
## Summary
- Adds all 8 ReUI-documented Gantt styling hooks — a comparison against ReUI's Gantt found none existed under any name: `data-selected`, `data-dragging`, `data-progress`, `data-completed`, `data-past`, `data-recurring` on the bar wrapper; `data-off` on the header date cell + off-day canvas band; `data-drop-invalid` on the JS drag ghost (renamed from the previously wrongly-named `data-invalid="true"`).
- New plumbing for the two hooks GanttBar had no state for: `Selected`/`Now` parameters on `GanttBar`, a new `GanttTimeline.SelectedIds` parameter wired to `GanttChart`'s existing `GanttState.SelectedIds`, and `data-dragging` set/cleared directly on the bar element by gantt-v3.js's pointer-driven drag engine (no Blazor round trip mid-gesture to hook into otherwise).
- `data-drop-invalid` now paints the ghost invalid immediately when a brand-new drag position starts validating, before the async verdict resolves — an unresolved predicate reads as invalid rather than optimistically showing a stale prior verdict, matching the drag's existing fail-closed commit gate.
- `GanttTask.IsRecurring` — additive flag (same pattern as `ParentId`), since GanttV3 has no recurrence-expansion engine of its own; backs `data-recurring` for a consumer generating its own repeating occurrences.
- Docs: new "Styling hooks" section on the GanttChart page with a per-family CSS example.

## Test plan
- [x] `dotnet build Lumeo.slnx -warnaserror` — 0 errors/0 warnings
- [x] New bUnit suite `GanttV3StylingHooksTests` (10 tests) — disable-check performed: neutering the hook gates flipped 8/10 to failing with the predicted wrong values (`null` instead of `""`/`"40"`/`"100"`), restored and green again
- [x] Full filtered suite (`FullyQualifiedName!~Lumeo.Tests.E2E&Category!=EjectGateFull`) — 7701 + 136 passed, 0 failed
- [x] Local E2E `Gantt/`, `Smokes/`, `Scheduler/` against own servers (ports 5299/5301, never 5287) — 241 passed, 0 failed; specifically confirmed by name: the new `Data_dragging_*` and `Leaf_Checkbox_Toggle_Marks_The_Bar_Data_Selected` specs, and the renamed `CanDrop_*` ghost specs
- [x] `python scripts/docs-parity/check.py` — exit 0
- [x] Registry regen + docs copy + Algolia index — no drift once committed

## New Features
- Added recurring-task support to Gantt charts.
- Added styling hooks for progress, completion, past, recurring, selected, off-day, dragging, and invalid-drop states.
- Gantt bars now support explicit selection and current-time positioning.

## Bug Fixes
- Preserved recurring-task and parent information during task interactions.
- Improved drag-preview styling and invalid-drop indicators.
- Corrected past-task detection for end dates and milestones.

## Documentation
- Documented recurring tasks and available Gantt styling hooks.


Brain2k-0005:
@codex review

Brain2k-0005:
All five findings are fixed in `6ba5a45c` — thanks, the P1 was a real data-loss path.

- **P1 IsRecurring across the JS round trip** — merged in `ReplaceTask` and `JsOnTaskClick` alongside `ParentId`. This one was worse than the ParentId case it mirrors: the merged task is written back into `_tasks` and re-emitted via `TasksChanged`, so the cleared flag became the new source of truth.
- **IsRecurring in the task hashes** — added to both `GanttChart.ComputeTasksHash` and `Gantt`'s own.
- **data-past on the server clock** — `RefreshBrowserNowAsync`'s `if (!NowIndicator) return;` is gone. Two existing tests pinned "disabled pays no interop round trip"; that cost contract was written when the now-line was the clock's only consumer, so they now pin the value being available instead, with the reason recorded.
- **data-past vs the rendered endpoint** — compares against `(IsMilestone ? Start : End).Date.AddDays(1)`, matching `BarGeometry`.
- **Ghost progress hooks** — `data-progress`/`data-completed` now track the preview percentage.

Disable-check: reverting the three C# fixes fails exactly the four new tests that should fail and leaves the two "must stay absent/present" ones green. Full suite 7726 + 136 + 90 + 37 + 32, zero failures.

Brain2k-0005:
@codex review

Brain2k-0005:
Round 2 fixed in `d80ea706`.

**The alias finding was the important one, and it was worse than "a rename".** This file's own comment on master described `data-invalid` as *the stable hook (E2E selector / consumer override)* — an explicit promise to consumers — and `Lumeo.Scheduler` paints its own drag ghost with the identical attribute plus a `[data-scheduler-ghost][data-invalid]` rule. So the rename would have broken existing overrides **and** split a cross-component convention, inside a PR whose entire purpose is to *add* styling hooks. Both attributes are now set and cleared together; the alias keeps its original `"true"` value so `[data-invalid="true"]` selectors keep matching, and the E2E asserts that exact value.

**Pending-verdict gap** — correct, and my own comment argued the wrong way: it claimed a revisited key "has nothing new to hide behind a pessimistic repaint", which holds for a *settled* verdict but not an in-flight one. Now tracked with a `pendingKeys` set.

On the E2E you suggested for it: I did not add one. A deterministic pending window needs a delay knob on the test page, and `CanDrop` is a synchronous `Func`, so the only way to stall it is blocking the very circuit the test drives — flaky by construction. The opposite failure direction *is* covered: if `pendingKeys` ever leaked, the drag-back-to-valid spec would fail, and it asserts both attributes clear. Flagging the gap rather than claiming coverage I don't have.

**Below-threshold spec** — now asserts absence with the pointer still down.

Verification: build 0/0, full non-E2E suite green, Gantt drag-parity E2E 30/30 against a ServerHost on :5299, Gantt smokes 10/10 against docs on :5301.

Brain2k-0005:
@codex review

Brain2k-0005:
Round 3 fixed in `5f1e23bd`. Both findings were caused by my own round-1 fixes, and one of them was the more serious kind — I turned a wrong CSS class into a crash.

**MaxValue overflow** — `RenderedEndExclusive` used `.AddDays(1)`, which throws `ArgumentOutOfRangeException` at `DateTime.MaxValue.Date`. A task using that as a "no deadline" sentinel rendered fine *before* this PR and would have crashed on render after it. Now compares whole calendar days (`EffectiveNow.Date > RenderedEndInclusive`) — equivalent arithmetic, no next-day construction. Covered by a `[Theory]` over both branches; reverting to the `AddDays` form fails both cases.

**data-past on move/resize ghosts** — `refreshGhostPast` recomputes from the snapped candidate dates on every move, mirroring `GanttBar.IsPast` exactly (same last-rendered-day rule, milestone reads Start).

---

**Correction to my previous comment.** I reported "Gantt drag-parity E2E 30/30" for round 2. That run used stale Release binaries — I had built Debug and then run `-c Release --no-build`, so my new assertions were not in the tested assembly. The count never moved from 30 even after adding a test, which is what exposed it. Re-verified properly against a fresh Release build:

- drag parity **31/31** (the new test is discovered and passes; the `data-invalid` alias assertion genuinely runs now)
- all non-smoke Gantt E2E **153/153** against ServerHost on :5299
- Gantt smokes **10/10** against docs on :5301
- full non-E2E suite green across two consecutive runs

One transient failure appeared in a single `Lumeo.Tests` net10.0 run and did not reproduce in two full re-runs; I could not attribute it, so I am noting it rather than calling it fixed.

Brain2k-0005:
@codex review

Brain2k-0005:
Fixed in `f82fc0b1` — correct, and it was my own regression from the previous round.

`refreshGhostPast` read `new Date()`. `GanttBar` derives `data-past` from `GanttTimeline.Now` whenever a consumer supplies one, so on a historical or simulated timeline the ghost contradicted the very bar it was cloned from — I fixed clone-staleness and introduced clock-divergence in the same function.

The effective now-date now travels in the drag-options snapshot as `nowDate`, and the engine reads only that, with **no** fallback to `new Date()`: if .NET sent nothing, the cloned attribute is left alone rather than repainted from a clock the bars are not using.

Date part only, deliberately — it is all the whole-day past-ness rule needs, and it keeps the field from changing on every clock resolve, so adding it to the equality-compared `GanttInteropOptions` does not turn the registration short-circuit into a re-registration treadmill.

Covered by a bUnit test asserting the options carry a **consumer-supplied** `Now` (2026-01-04, deliberately not today) rather than the ambient clock.

Verification: Release build 0/0, full non-E2E suite green, all non-smoke Gantt E2E **153/153**.

Brain2k-0005:
@codex review

Brain2k-0005:
Both fixed in `39d5a3a8`.

**Leaf-only `data-selected`** — gated on the row not being a hierarchy parent. Worth flagging how I got the parent test: `row.Rollup` looks like a ready-made "is this a parent" check and is a trap — `Rollups` is empty whenever `ShowSummaryBars` is off, so keying the hook off it would have made `data-selected` flip with an unrelated display toggle. Parent-ness is now stamped once per render onto the row record, built from the same `FilterValidDurationTasks` set the row builder uses — the same filtering the round-2 P2 #7 fix on `usesHierarchy` established directly above it. A test asserts the gate holds for **both** values of `ShowSummaryBars`.

**Standalone timeline clock** — `GanttTimeline` with no host and no `Now` now resolves the browser clock itself rather than falling back to the server's `DateTime.Now`. Self-gated to "no `Now` supplied", so the `GanttChart`-hosted path pays no second round trip, and a controlled `Now` still wins outright. The call sits after the mount-scroll claim and before the registration awaits, so it does not reopen the #390 claim-then-await shape.

---

Unrelated but worth recording, because it cost real time and would cost it again: my first attempt put the explanatory Razor comment **inside** `<GanttBar>`'s attribute list. That compiles with **0 errors and 0 warnings** and then misbinds parameters at runtime — `data-selected` appeared on bars with an empty `SelectedIds`, and two tests that had been green all PR started failing. The tell was that tests unrelated to the logic change flipped too. Comment moved above the tag.

Disable-check: removing the leaf gate fails exactly the two new tests and nothing else. Release build 0/0, full non-E2E suite green, non-smoke Gantt E2E **153/153**.

Brain2k-0005:
@codex review

Brain2k-0005:
All three fixed in `b4074205`. Two of them correct a claim I made in this thread, so let me be explicit.

**Drag registration was delayed — including on the hosted path, which I said it wouldn't be.** I wrote that `ResolveOwnBrowserNowAsync` self-gates to "no `Now` supplied" and therefore costs the `GanttChart` composition nothing. That is false on the first render: `GanttChart` passes `Now="@_browserNow"`, which is null until its own resolve lands, so the gate did not hold and the extra round trip ran — ahead of `SyncDragRegistrationAsync`, in a section whose own comments say drag registration must be the first round trip. Moved to last among the mount awaits; nothing about `data-past` is interaction-critical.

**Cleared `Now` was never re-resolved.** Only call site was the `firstRender` branch, so a timeline that mounted with a `Now` and had it cleared kept the server clock for the rest of the session. Now also attempted on later renders, self-gated so it stays a no-op otherwise.

**Parent-ness came from the wrong source.** `Rows` overrides `Tasks`, so rebuilding topology from `Tasks` stamped a declared parent row as a leaf. Worth recording what the fix turned up: switching to `row.HasChildren` alone broke my two earlier tests, because `BuildFallbackRows` — the `Rows == null` path — hardcodes `HasChildren: false` for **every** row, so a standalone timeline fed through `Tasks` reports every parent as a leaf. Neither source alone covers both modes, so the gate now reads `row.HasChildren || <ParentId links among the rendered rows>`, both taken from the authoritative row list. New test covers the `Rows`-supplied case with `Tasks` deliberately empty.

Release build 0/0, full non-E2E suite green, non-smoke Gantt E2E **153/153**.

Brain2k-0005:
@codex review

PR #407, opened 2026-08-12
feat(scheduler): add ISO week numbers to the first-party month grid

Brain2k-0005:
Closes a real capability gap found by comparing Lumeo's first-party event calendar against **reui.io/components/event-calendar**, which lists a *week numbers* toggle under its display controls. Lumeo's month grid had no way to show them at all — a consumer could not do this, with any amount of CSS.

## Design decision worth reviewing

The label is the **ISO-8601 week read off each row's own Thursday**, not off the row's first cell.

Thursday is ISO's defining day (the week a Thursday falls in *is* its ISO week), and every 7-day row contains exactly one Thursday no matter which `FirstDayOfWeek` the grid starts on. Reading the row's first cell looks equivalent and is not: in a **Sunday-start** grid the leading Sunday belongs to the *previous* ISO week, so that reading labels most rows one week too low.

A test pins exactly this — a Monday-start and a Sunday-start grid must agree on every row covering the same Thursday. Disable-check: swapping the implementation to the naive first-cell reading fails that test and only that test; the other four stay green.

## Non-goals / safety

- `ShowWeekNumbers` defaults to **false**, and with it off the grid renders its **exact previous seven-column markup** — pinned by its own test, so no existing consumer's layout can shift.
- The column is a `rowheader`, not a `gridcell`: the month grid's arrow-key navigation walks day cells by index, and injecting a non-focusable gridcell would have broken that. A test asserts the grid still has exactly 42 gridcells.
- Grid class strings are literal (`grid-cols-7` / an explicit template), never interpolated — Tailwind v4 only emits utilities it can see in source, and this repo has been bitten by a class in the DOM with no rule behind it.

## Verification

Release build 0 errors / 0 warnings, full non-E2E suite green, docs-parity exit 0, registry regenerated.

## Remaining ReUI gaps (not in this PR)

For the record, from the same comparison: **resource lanes** (`SchedulerEvent.ResourceId` already exists in the data model, but no view groups by it), **live-region announcements** (there is no `aria-live` anywhere in the views), **weekend hiding**, **raw RRULE string parsing** (the structured `SchedulerRecurrenceRule` covers FREQ/INTERVAL/COUNT/UNTIL/BYDAY, but iCalendar strings cannot be passed in), and **IANA time zones** (deliberately absent today — the kernel is documented as wall-clock-only).


Brain2k-0005:
@codex review

Brain2k-0005:
All three fixed in `a0ddf510`. The first one is the embarrassing one, so let me name it plainly.

**The CSS class had no rule.** My PR text explicitly cited the class-in-the-DOM-without-a-rule trap and said the grid classes are written literally so Tailwind can see them — and then I did not regenerate the bundles, which predate the class. Writing it literally is *necessary but not sufficient*. Both bundles are rebuilt now (`npm run build:css` and the docs project's own `css:build`), and I verified the rule landed with a literal substring scan in Python rather than grepping a minified single-line file.

**Header and body columns did not line up.** Correct — they are two separate grids, each resolving `auto` independently against its own content, and the two cells had different padding (`px-2` vs `px-1`). Swapped to a fixed `2.5rem` track, which resolves identically in both, and matched the padding.

**`aria-hidden` on the Wk/KW header was the wrong trade.** I reasoned it would avoid an empty-column-header announcement, but the cell is not empty — it carries the only text that explains what the column is, so hiding it left the row headers as bare numbers like "35" with nothing to relate them to. Now exposed, with a test that positively asserts the labelled element exists and carries neither `aria-hidden` nor `role="presentation"` (rather than merely asserting a bad string is absent).

Release build 0/0, full non-E2E suite green.

Brain2k-0005:
@codex review

Brain2k-0005:
Both findings closed in `ab72d270`.

**Locale coverage** — correct, and the gap was wider than one key: only English and German had `Scheduler.WeekNumberAbbrev`, so under any other supported culture the column header read `Wk` while everything around it was translated. Added to all 12 remaining locale files, and — more importantly — added to `P2LocalizationSweepTests.NewKeys`, so a future omission fails the gate instead of degrading silently at runtime. That guard is the part worth keeping; the translations alone would have gone stale the next time someone added a key.

**Stale test count** — resolved by regenerating; the manifest now reflects the current `[Fact]`/`[Theory]` total.

Clean Release build 0 errors / 0 warnings, full non-E2E suite green.

Brain2k-0005:
@codex review

Brain2k-0005:
Both fixed in `ad58b1ac`.

**The ARIA finding was right and my previous fix was incomplete.** Removing `aria-hidden` made the `Wk`/`KW` label visible, but that label sits in the header strip *above* the grid — it is not a `columnheader` of the grid and has no ARIA relationship to the row headers, so a screen reader walking the grid still reached an unexplained "35". Each row header now carries its own localized `aria-label` ("Week 35" / "Kalenderwoche 35"), which holds regardless of what the header strip does. Translated across all 14 locales and added to the localization sweep. The test asserts the label contains the number *and* differs from it, so a bare number cannot pass.

**Docs** — `ShowWeekNumbers` is now in the `SchedulerMonthView` parameter table, which is where consumers discover these views.

One thing I want on the record rather than buried: during verification this branch showed a red run with two failures — `GanttV3KeyboardNavigationTests.Mid_Flight_Combination_Drift` and `ToastStackingTests.Focus_Moving_Between_Two_Controls_In_The_Group_Never_Collapses_It`. Neither is code this branch touches. Three subsequent runs here were green and three runs on unmodified `master` were green. Both tests poll in 10 ms steps, so they look load-sensitive rather than broken — but I could not reproduce it deterministically, so I am flagging it rather than calling it explained.

Brain2k-0005:
@codex review

PR #408, opened 2026-08-12
feat(scheduler): announce drag outcomes in a live region

Brain2k-0005:
Second gap closed from the **reui.io/components/event-calendar** comparison: its accessibility section lists *live-region announcements for gestures*. Lumeo's first-party views had **no `aria-live` region anywhere** — a pointer drag produced no screen-reader signal at all.

> **Stacked on #407** (both touch `SchedulerMonthView`). Base is `feat/scheduler-week-numbers`; retarget to `master` once that merges.

## What is announced, and why these choices

- **The rejected drag matters most.** When `CanDrop` refuses, the event visibly springs back — a purely visual cue. That path previously returned silently. It now announces, and deliberately does **not** name the attempted date, so it cannot be misheard as a successful move.
- **Announced after any `CanDrop` adjustment**, from the committed values. An adjusted drop lands somewhere other than where the pointer was released, so announcing the *proposed* date would state something that did not happen. A test pins this.
- **Month announces the date, TimeGrid the start and end times** — times are the whole point of a time-grid gesture, and that wording covers resize as well as move.
- **Long date format, not the cell label.** Read aloud without the grid for context, "12" says nothing; "12 March 2026" does.
- **The region is rendered empty at mount**, not created when the first announcement happens: assistive tech must already be observing the node when the text lands. Same `sr-only` / polite / atomic shape `Carousel.razor` already uses — house pattern, not a second invention.

## Verification

Every test drives the real `[JSInvokable]` commit path the drag engine calls, never a field or parameter, so a regression that drops the announce call from that path cannot pass. Disable-check: removing both month announces fails exactly the three behavioural tests and leaves the two structural ones green.

Release build 0 errors / 0 warnings, full non-E2E suite green, docs-parity exit 0.

## Still open from the same comparison

**Resource lanes** (`SchedulerEvent.ResourceId` exists in the data model; no view groups by it), **weekend hiding**, **raw RRULE string parsing**, **IANA time zones** (deliberately absent — the kernel is documented wall-clock-only).


Brain2k-0005:
@codex review

Brain2k-0005:
All six fixed in `6814a32b`. The P1 is the one that matters, and it invalidated this PR's headline claim.

**The rejection announcement never fired.** I put it in `CommitDrag`, which an ordinary `CanDrop` refusal never reaches — the JS awaits `ValidateDrop` and returns at `scheduler-views.js:190-196` (and 456-463) without ever invoking it. So the case I opened this PR by calling *"the one that matters most"* was precisely the case that did nothing.

Worse, **my own test hid it**: it called `CommitDrag` directly, which felt like "driving the real commit path" and is not the path the engine takes for a rejection. Testing the invokable is not the same as testing the gesture. Both abort sites now call a `NotifyDropRejected` invokable, and the tests drive that seam instead.

The rest, all correct:
- **Resize vs move wording** — the time-grid branch handles `ResizeStart`/`ResizeEnd` too, so the mode now travels with the rejection and a refused resize is not announced as a refused move.
- **No announcement without `OnEventChange`** — the drag is ghost-only in that case, the chip snaps back and `Events` is untouched, so "moved to …" described something the user could neither see nor keep. That is live on the docs Day view today.
- **Repeated identical messages** — assigning the same string mutates no text node. Worth recording that my first attempt at this was wrong too: clearing and re-setting inside one method does nothing, because Blazor coalesces the two `StateHasChanged` calls into a single render. The region is now `@key`'d on an epoch, so a repeat genuinely replaces the element.
- **Locales** — all five keys are translated across the 12 remaining locale files, and added to `P2LocalizationSweepTests.NewKeys` so a future gap fails the gate rather than silently falling back to English.

Not done, and I would rather say so than let it look covered: **drag-created ranges are still unannounced** (`CommitDateRangeSelect` / `CommitCreate`). It is a genuine gap, but it is a different gesture with its own wording and test surface, and folding it in here would widen a PR that has already changed shape twice.

One process note: a clean rebuild caught ten analyzer errors (`xUnit1031`) that the incremental build reported as zero. `-warnaserror` on an incremental build is not the gate it looks like.

Clean Release build 0/0, full non-E2E suite green (7742 per framework).

Brain2k-0005:
@codex review

Brain2k-0005:
Superseded by #410 — GitHub auto-closed this when `feat/scheduler-week-numbers` was deleted on merging #407, and a closed PR cannot be reopened or retargeted. Same branch, same work, now against `master`. The review history that shaped it stays here.

PR #409, opened 2026-08-12
feat(scheduler): parse RFC 5545 RRULE strings

Brain2k-0005:
Third gap closed from the **reui.io/components/event-calendar** comparison: it accepts raw RRULE strings. Lumeo modelled recurrence only as a structured `SchedulerRecurrenceRule`, so anyone holding iCalendar / Google / Outlook data — where recurrence *is* a string — had to write a parser before they could pass their own data in at all.

## The design decision worth reviewing: strict, not lenient

Scope is exactly the subset `SchedulerRecurrenceExpander` implements — `FREQ=DAILY|WEEKLY|MONTHLY`, plus `INTERVAL`, `COUNT`, `UNTIL`, `BYDAY` with ordinals (`2MO`, `-1FR`). Anything else — `FREQ=YEARLY`, `BYMONTHDAY`, `BYSETPOS`, `WKST` — is **rejected**, not silently dropped.

That is the whole point. A lenient parser would accept `FREQ=MONTHLY;BYMONTHDAY=15` and return a monthly rule that ignores the day constraint: the caller gets a series that looks right, is wrong, and is built from their real calendar. A rejected parse is a bug report on day one; a silently narrowed rule is a support ticket six months later. Fifteen of the tests pin exactly this.

Also deliberate: `UNTIL=...Z` is accepted but **not** converted to local time. This kernel is documented wall-clock throughout — it never calls `ToLocalTime`/`ToUniversalTime` — so shifting only that one value would make `UNTIL` disagree with every other date in the same rule.

## Verification

The structural assertions are the weaker half; the real proof is two end-to-end tests that run the parser's output through the expander it exists to feed, because a structurally plausible rule the engine cannot use would pass everything else. Their expected dates were checked against the calendar independently, not read off the engine: `FREQ=WEEKLY;BYDAY=MO,WE;COUNT=4` → 10/12/17/19 Aug 2026, and `FREQ=MONTHLY;BYDAY=-1FR;COUNT=3` → 28 Aug / 25 Sep / 30 Oct 2026, each genuinely the last Friday of its month.

Release build 0 errors / 0 warnings, full non-E2E suite green, docs-parity exit 0.

## Remaining from the comparison

**Resource lanes** (`SchedulerEvent.ResourceId` exists in the data model; no view groups by it) and **weekend hiding**. IANA time zones I would not call a gap — the kernel is deliberately wall-clock, and changing that is an architecture decision rather than an addition.

## Summary by CodeRabbit

* **New Features**
  * Added RFC 5545 recurrence-rule parsing for daily, weekly, and monthly schedules.
  * Supports intervals, counts, end dates, weekday selections, optional prefixes, and flexible capitalization and spacing.
  * Added safe validation and exception-based parsing options.

* **Bug Fixes**
  * Invalid, unsupported, duplicate, and incompatible recurrence rules are now rejected consistently.

* **Tests**
  * Added comprehensive coverage for valid formats, edge cases, and recurrence expansion.


Brain2k-0005:
@codex review

Brain2k-0005:
All six fixed in `f7a813bb`. The P1 is the one that stings, so let me name it.

**I rejected unknown part *names* but not unhonourable *combinations* — which is the exact failure this PR's own description argues against.** Three inputs parsed cleanly and then meant something else:

- `FREQ=DAILY;BYDAY=MO` → every day, because daily expansion ignores `ByDay`.
- `FREQ=WEEKLY;BYDAY=2MO` → every Monday, because weekly expansion ignores ordinals.
- `FREQ=MONTHLY;BYDAY=MO` → only the first Monday, while RFC 5545 reads an unqualified monthly term as *every* Monday — a divergence `SchedulerByDayRule.Ordinal` already documents as out of scope, which makes accepting it a silent narrowing rather than an oversight in the model.

All three are rejected now, in one `ByDayIsHonouredBy` gate applied before the rule is constructed.

The rest, all correct and all the same shape:
- **Ordinal range** — `54MO` parsed, matched no date in any month, and yielded a silently empty series. Bounded to ±1..53.
- **Duplicate `BYDAY` terms** — `1MO,1MO` reached monthly expansion as two rules, emitting two instances at the same timestamp and burning the whole `COUNT` on one date. Rejected, consistent with how duplicate parts are already handled.
- **Duplicate `INTERVAL`** — I deduplicated `FREQ`/`COUNT`/`UNTIL`/`BYDAY` and missed `INTERVAL`, so the last value silently won.
- **Empty segments** — `RemoveEmptyEntries` swallowed `;FREQ=DAILY`, `FREQ=DAILY;` and `FREQ=DAILY;;COUNT=2`, which is precisely the truncated calendar line the strictness exists to surface.

**And a correction to my own comment.** I wrote that the DATE vs DATE-TIME distinction for `UNTIL` was "moot". It is not: `Until` is bounded by date, so `UNTIL=20260812T080000` still admitted that day's 09:00 occurrence — later than the caller asked for. A DATE-TIME is now accepted only at `23:59:59`, the end-of-day idiom real calendars emit and the one time-of-day where date-only bounding is exactly equivalent. Everything else is rejected rather than silently widened to the full day.

Twelve new negative cases pin these. Release build 0/0, full non-E2E suite green.

Brain2k-0005:
@codex review

Brain2k-0005:
All six fixed in `2cab9a73`. Two are worth spelling out.

**The packaging finding was right, and the fix is a move rather than a manifest edit.** `RegistryGen` builds a component's vendored `files` list from its own `UI/<Name>/` folder, so **nothing** under `Kernel/` reaches a source-vendored install — not this parser, and not the expander either. Special-casing one file would have papered over that. The parser now lives in `UI/Scheduler/`, beside `SchedulerRecurrenceRule.cs`, the type it produces — which is where it belonged anyway. Verified by regenerating: `files` now lists `UI/Scheduler/SchedulerRRuleParser.cs`.

**The ordinal range was wrong in a way my own reasoning should have caught.** I bounded it to RFC 5545's ±1..53, but that range exists for `YEARLY`, which this subset does not implement. Here ordinals are only honoured for `MONTHLY`, and no month has a sixth occurrence of a weekday — so `6MO` resolved to nothing every month and the expander spun to its safety cap and returned a silently empty series. Bounded to ±1..5.

The rest:
- **`Math.Abs(int.MinValue)` throws**, so `BYDAY=-2147483648MO` made a documented never-throwing `TryParse` throw. Compared with a range pattern instead.
- **Mixed-sign ordinals for one weekday** — `1MO,-5MO` are distinct terms that resolve to the same date in any five-Monday month, duplicating an instance and burning `COUNT`. Same failure the exact-duplicate check catches, spelled differently.
- **Nested `BYDAY` split** still dropped empty entries after I fixed the outer one, so `MO,`, `,MO` and `MO,,WE` parsed.
- **`UNTIL=20261231Z`** — `Z` belongs to the DATE-TIME form; stripping it unconditionally made a malformed date-only value parse.

Eight more negative cases pin these; 52 tests total. Clean Release build 0/0, full non-E2E suite green, docs-parity exit 0.

Brain2k-0005:
@codex review

Brain2k-0005:
All three fixed in `8aadddf6`. The first one corrects **my** over-correction.

**Mixed-sign ordinals: I was too strict.** I refused any positive/negative pair for one weekday. But a weekday occurs 4 or 5 times in a month, so a positive `n` and a negative `-j` land on the same date exactly when `n + j` is 5 or 6. `1MO,-1MO` — first and last Monday — sums to 2 and can never coincide; the expander emits both correctly, and I was rejecting a faithful rule. Now refused only for sums of 5 or 6, with a behavioural test that expands `1MO,-1MO` and checks the dates against the calendar (March 2026 has Mondays on 2, 9, 16, 23, 30 → first 2, last 30).

**INTERVAL** is capped at 10 000. `DailyCandidates` calls `AddDays(interval)` before `ExpandRule` can notice `COUNT` is satisfied, so a near-`MaxValue` interval threw out of a rule the parser had just declared valid. 10 000 is 27 years of daily steps — far past anything real, and it cannot overflow.

**COUNT** above the expander's 500-candidate cap is refused: it would otherwise return fewer occurrences than requested, which is precisely the accept-then-quietly-do-something-else outcome this parser exists to prevent.

Clean Release build 0/0, full non-E2E suite green across three consecutive runs.

Brain2k-0005:
@codex review

Brain2k-0005:
Three fixed in `8f24a1a6`; on the P1 I am pushing back in part.

**The unvendored dependency was the sharp one, and it was self-inflicted.** Moving the file into `UI/Scheduler/` fixed the packaging gap — and then my `COUNT` check referenced `SchedulerRecurrenceExpander.OccurrenceCap`, recreating the exact same problem as a *type* reference: it compiles in this repo and would not compile in a source-vendored install. The cap is a local literal now, and the round-trip tests (parser output → real expander) are what keep it honest. Zero `SchedulerKernel` references remain in the file.

**INTERVAL** — a single flat cap was wrong, as you say. The expander takes up to 501 steps and advances *before* it can notice `COUNT` is satisfied, so the safe bound depends on how far one step moves: ~5 700 daily, ~820 weekly, ~190 monthly to stay inside `DateTime`. Bounded per frequency at 5 000 / 800 / 150, enforced after the loop since `FREQ` may appear after `INTERVAL` in the string.

**UNTIL case** — correct, and inconsistent of me: I lowercased every other token but matched `T` as a literal, so `until=20261231t235959z` was refused while `freq=daily` was fine.

---

**On the P1 — bare `FREQ=MONTHLY` with a month-end `DTSTART` — I disagree with rejecting, and have documented it instead.**

The divergence is real: RFC 5545 skips months lacking the anchor day, the expander clamps to the last day (its own documented wave-1 choice), so importing a 31st-anchored rule yields a February date the source calendar does not have.

But it is invisible from the rule text. It depends entirely on `DTSTART`, which this parser never sees. Rejecting every bare monthly rule to catch it would refuse the *overwhelming majority* — anchors on days 1–28, where the two readings are byte-identical — and make importing ordinary monthly events impossible. That trades a narrow, data-dependent inaccuracy for a broad, unconditional failure.

The check belongs where rule and `DTSTART` are both known, which is the expansion call, not here. The class docs now state the divergence explicitly and point month-end importers at an explicit `BYDAY` ordinal, which resolves exactly. Happy to be overruled — if you would rather refuse them outright, say so and I will make it strict.

61 tests. Clean Release build 0/0; full non-E2E suite green across three consecutive runs.

Brain2k-0005:
@codex review

Brain2k-0005:
Both fixed in `9bca1b4f`, and the P1 deserves a plain statement rather than a fix note.

**The P1 is correct, and I verified it directly.** `Scheduler.razor`'s `ToJsEvent` branches only on `DaysOfWeek`; `ev.Recurrence` appears exactly once in that file — inside the events *hash*, for change detection — and never in the serialized payload. So a caller who parses an RRULE, assigns it to `SchedulerEvent.Recurrence` and renders `<Scheduler>` gets **nothing**. It works only through the first-party views, which expand it via the expander.

That is a pre-existing limitation of the `Recurrence` property, not something this parser introduced — but the property's own XML docs make it worse: they say it "takes precedence over `DaysOfWeek`" and describe the expander, without ever saying the FullCalendar wrapper ignores it. I have documented the boundary on the parser so nobody discovers it by watching an imported series not appear.

I have **not** made `<Scheduler>` honour it here. That is a real feature — expand occurrences server-side and feed FullCalendar concrete events, or adopt its premium rrule plugin — and it is a design decision, not a parser change. Worth its own issue.

**Monthly INTERVAL** — right again, and my arithmetic was too generous: the cap counts *yielded* candidates while `MonthlyCandidates` advances its offset even for months that yield none, so `BYDAY=5MO` steps roughly three times as many months as it produces. Dropped from 150 to 50.

Brain2k-0005:
@codex review

Brain2k-0005:
Both fixed in `6b6e05b7`, and the UNTIL one reverses a decision I defended earlier in this thread.

**UTC `UNTIL` is now rejected outright.** I had argued for accepting `Z` unconverted because the kernel is wall-clock. Your example shows why that is wrong in a way my reasoning skipped: in UTC+10, `UNTIL=20260812T235959Z` is 2026-08-13 09:59:59 local, so a daily 09:00 occurrence on the 13th is *inside* the cutoff — and parsing the date part alone yields the 12th and drops it. Not a rounding difference; a whole occurrence, silently. The parser never sees the target zone, so it cannot do better than refuse.

The cost is real and I want it on the record: `...T235959Z` is the form most calendars emit, so importers must convert to local wall-clock before parsing. That is worse ergonomics for correct data, which is the same trade this parser makes everywhere else — and it would have been inconsistent to accept a silent day-shift here while rejecting `BYMONTHDAY` for being unhonoured.

**Monthly INTERVAL** → 12. `INTERVAL=48;BYDAY=5TH` samples one month every four years, so it can walk past `DateTime`'s range long before the occurrence cap fires. A year per step is ample for any real calendar.

Clean Release build 0/0, full non-E2E suite green, 62 tests.

Brain2k-0005:
@codex review

PR #410, opened 2026-08-13
feat(scheduler): announce drag outcomes in a live region

Brain2k-0005:
Replaces #408, which GitHub auto-closed when its base branch `feat/scheduler-week-numbers` was deleted on merging #407. Same branch, same work, now targeting `master` — the full review history lives on #408.

Second gap closed from the **reui.io/components/event-calendar** comparison: its accessibility section lists *live-region announcements for gestures*. Lumeo's first-party views had **no `aria-live` region anywhere**, so a pointer drag produced no screen-reader signal at all.

## What it announces, and the three corrections that shaped it

- **The rejected drag is the case that matters** — the event visibly springs back, which is a purely visual cue. My first attempt announced from `CommitDrag`, which an ordinary `CanDrop` refusal **never reaches**: the JS awaits `ValidateDrop` and returns first. The headline feature did nothing, and my own test hid it by calling `CommitDrag` directly — driving a path the engine does not take for rejections. Both JS abort sites now call a `NotifyDropRejected` invokable.
- **Commit-time rejections are announced too.** That branch is unreachable for an ordinary refusal but *is* the path for the staleness case it exists to catch (`CanDrop` said yes during the drag, no at the drop), which would otherwise fail silently.
- **Repeating an identical message still mutates the DOM.** Two earlier attempts were wrong: clearing and re-setting inside one method does nothing (Blazor coalesces the renders), and keying the element on an epoch mutates the DOM by *replacing the observed node* — which can suppress announcements entirely in readers that only watch an existing live region. The node stays mounted; only its text changes, alternating a named zero-width space.
- **Announced after any `CanDrop` adjustment**, from committed values, and **not at all** when `OnEventChange` has no delegate — the drag is ghost-only then, so "moved to …" would describe something the user can neither see nor keep.
- Month announces the date, TimeGrid the start and end times, and a refused **resize** is not announced as a refused move.

All five keys are translated across the 14 shipped locales and added to `P2LocalizationSweepTests`, so a future gap fails the gate instead of falling back to English.

## Verification

Every test drives the real `[JSInvokable]` seam the drag engine calls. Clean Release build 0 errors / 0 warnings, full non-E2E suite green.

Merging master into this branch conflicted across 16 source files; resolved per hunk rather than wholesale, and verified afterwards that **both** #407's `aria-label` work and this PR's announcements survived — a blanket `--ours` had silently dropped the former on the first attempt.

## Still open from the same comparison

**Resource lanes** and **weekend hiding**. Drag-*created* ranges are also not announced yet — a separate gesture with its own wording and test surface.


Brain2k-0005:
@codex review

Brain2k-0005:
Both fixed in `bc25bfe4`.

**The out-of-bounds drop was the better catch.** I had routed rejections through `NotifyDropRejected`, but that call sits *inside* the `hasCanDrop` branch — so a time-grid release outside every `[data-daycol]`, with no `CanDrop` configured, skipped validation entirely, `CommitDrag` returned on the null day, and the gesture ended in silence. Exactly the failure this PR exists to remove, in the one configuration I had not walked. The null-day case is now reported before that branch.

**No-op adjustments** — correct: `CanDrop` can accept *with* an adjustment that snaps the event back to where it started, and the earlier no-op check compared the **proposed** values, not the adjusted ones. So it announced a move for an event whose committed state is unchanged. Both views now compare the committed values against the source before announcing, with a test that drives an adjustment back to the original start and end.

Clean Release build 0/0, full non-E2E suite green.

Brain2k-0005:
@codex review

Brain2k-0005:
Both fixed in `4ef151a4`. The P1 was **my regression from the previous round**, and it was a functional break, not a cosmetic one.

I added the missing-day guard unconditionally. But `onPointerMove` populates `dayIso` **only** for `mode === 'move'` and deliberately leaves it null for both resize modes — so every resize was treated as a drop outside the grid: it announced a rejection and returned before committing. **Resizing stopped working entirely.** Scoped to moves now.

Worth naming the pattern, because it is the third time in this PR: each fix I made to close a hole opened a new one one layer down — announce-in-the-wrong-method, then a `@key` that could suppress all announcements, now a guard that killed resize. Every one was caught here rather than by me, and each is a case where the change looked local and was not.

**Announce after the await** — correct. An async `OnEventChange` that faults left the user hearing a successful move while the JS caller quietly caught the failed `CommitDrag`. A test now throws from the handler and asserts the region stays empty.

Clean Release build 0 errors / 0 warnings, full non-E2E suite green.

Brain2k-0005:
@codex review

PR #411, opened 2026-08-13
feat(scheduler): add weekend hiding to the first-party month grid

Brain2k-0005:
Fourth gap closed from the **reui.io/components/event-calendar** comparison — its display controls include a weekend-hiding toggle, and Lumeo's month grid had no way to drop those columns at all.

## The rendering is the easy half; the keyboard is the risk

Arrow navigation moved by **raw 0–41 cell index** (`±1` for a day, `±7` for a week). Simply not rendering two columns would have left the arrow keys targeting cells that are **not in the DOM** — focus would go nowhere, silently, and only for keyboard users.

Navigation now walks positions in the *visible-cell* list. With weekends shown that list is the identity `0..41`, so the existing path keeps its exact previous arithmetic; with them hidden, the hidden columns are stepped over. `Home`/`End` snap to the row's first and last **visible** column, so `End` lands on Friday rather than a Sunday that isn't there.

Disable-check: reverting just the arrow-right and End cases to raw-index stepping fails exactly those two specs and nothing else.

## The other things that would have broken it

- **The weekday header strip drops the same two columns.** A 5-column body under a 7-column header shears the whole grid; a test counts the strip's own children.
- **All four grid-template combinations are literal class strings**, and both shipped CSS bundles are rebuilt. On #407 I wrote the class literally, did *not* rebuild the bundles, and shipped a class with no rule behind it — verified here by scanning both minified bundles for the new template.
- **Composes with `ShowWeekNumbers`** — pinned by its own test, since the two together are the combination most likely to be forgotten.

## Non-goals

`HideWeekends` defaults to **false**, and with it off the grid renders 42 cells and `grid-cols-7` exactly as before — asserted directly, so no existing consumer's layout can shift. The 42-cell model is untouched; only rendering and navigation skip columns.

## Verification

Clean Release build 0 errors / 0 warnings, full non-E2E suite green, docs-parity exit 0. `ShowWeekNumbers` and `HideWeekends` are both listed in the docs parameter table.

## Remaining from the comparison

**Resource lanes** — the last one, and the largest: `SchedulerEvent.ResourceId` exists in the data model, but no view groups by it, so it is a new view rather than a toggle.


Brain2k-0005:
@codex review

Brain2k-0005:
All three fixed in `ac728db2`. Each was a real hole in my implementation, and the disable-check turned up a fourth problem in my own test.

**P1 — no focusable cell.** Correct and worse than it sounds: the roving-tabindex anchor defaults to index 0, which in a **Sunday-first** grid is a hidden Sunday, so *every* cell sat at `tabindex="-1"`. The grid fell out of the tab order entirely and arrow keys did nothing — keyboard users lost the component, silently. Toggling the option on while a weekend cell held focus stranded it the same way. The anchor is now moved onto a rendered cell, with tests for both routes.

**Titles on runs starting in a hidden weekend.** `ShowTitle` is `!contL` — only a run's first day carries the name — so an event starting Saturday and continuing into the week lost its only titled chip and rendered as an unlabelled bar. The title now moves to the first *visible* segment.

**Vertical moves at the edges.** Right: clamping a flattened position changes its column, so `ArrowUp` in the top row slid from Friday to Monday. Edge moves now stay put.

---

Worth recording: the disable-check caught that **my title test was worthless**. Reverting the fix left it green, because `Assert.Contains("Offsite", cut.Markup)` also matches the tooltip and the `aria-label` — the title is in the markup even when every visible chip renders a blank. It now asserts the chip's own `TextContent`, and fails against the reverted fix as it should.

And a process note that cost me twice today: an **incremental** `-warnaserror` build reported `0 Error(s)` while a clean one surfaced an `xUnit2031` analyzer error in this PR's own new test. Analyzers do not re-run on inputs that look unchanged, so only a clean build is a gate — and my check now evaluates the build result as well as the test result, rather than printing a fixed string.

Clean Release build 0/0, full non-E2E suite green, docs-parity exit 0.

Brain2k-0005:
@codex review

PR #412, opened 2026-08-13
feat(scheduler): add a first-party resource view

Brain2k-0005:
**The last of the five gaps** from the reui.io/components/event-calendar comparison. ReUI renders *one column per bookable resource*, activated by supplying resources.

Lumeo already had `SchedulerResource` and `SchedulerEvent.ResourceId` — but they only drove **colour-coding and a legend**. No view ever grouped *by* resource, so a room-booking or per-person schedule could not be expressed at all. The model was there; the view was not.

## Design decisions worth reviewing

**A separate component, not a mode of `SchedulerTimeGridView`.** That view's columns *are* dates throughout — `_dayColumns`, `data-daycol`, and a drag/create engine that keys every gesture off the day under the pointer. Making a column sometimes mean "resource" would fork the most delicate code in the interaction layer. What would genuinely have been duplicated — the time axis and the overlap packer — are pure kernel calls, and both are reused.

**Overlaps pack within a column, never across.** Two rooms booked at the same hour is the *point* of the view, not a conflict; side-by-siding them across columns would invert its meaning. Pinned by a test that asserts two same-time bookings on different resources each keep full width, while two on one resource each take half.

**An event whose `ResourceId` matches no supplied resource is not rendered.** Folding it into the first column would misattribute someone's booking, and inventing an "unassigned" column is a layout decision this view has no basis to make on the caller's behalf. Documented on `Resources` and pinned by a test, so it is a stated contract rather than a silent drop.

**Recurring bookings** go through the same expander the month and time grids use, so they appear here exactly as they do there.

## Scope, stated plainly

**Read-only in this first version — no drag, resize or create.** Those live in the time grid's pointer engine, and wiring it to resource columns is separate work. A half-wired gesture would be worse than none, and this PR has enough surface as it is.

## Verification

Clean Release build 0 errors / 0 warnings, full non-E2E suite green, docs-parity exit 0. Disable-check: removing the per-resource filter fails exactly the three tests that depend on it. The grid label is translated across all 14 locales and added to the localization sweep.

## Summary by CodeRabbit

- **New Features**
  - Added a read-only, single-day scheduler view organized by resources.
  - Displays all-day and timed events with recurrence support, overlap handling, filtering, scrolling, and configurable time ranges.
  - Supports event selection, custom colors, accessibility labels, and empty-resource states.

- **Localization**
  - Added resource-grid labels across all supported languages.

- **Tests**
  - Expanded coverage for resource ordering, event placement, recurrence, accessibility, filtering, interaction, and time handling.


Brain2k-0005:
@codex review

Brain2k-0005:
All eleven fixed in `8483ec5c`. **Two were crashes**, and several were cases I had simply not walked for a brand-new component.

**Crashes.** Duplicate event ids produced identical instance keys and threw out of the packer's `ToDictionary`. A `SlotMinTime` of 23:30 pushed the minimum-span rule past midnight, and the resulting hour 24 threw out of `DateTime`'s constructor while formatting the axis.

**Correctness, each a case that silently misled rather than failed loudly:**
- **All-day bookings vanished** — the `!AllDay` filter dropped them entirely, and a room booked for a whole day is an ordinary case. They now have their own strip.
- **Overnight bookings** — the expander returns a 22:00–02:00 reservation whole for *both* days it touches, so the second day drew it from 22:00 again. Clipped to the day in view.
- **Bookings outside the visible hours** were clamped into a sliver at the top, implying a reservation that is not shown. Skipped instead.
- **Fractional `SlotMinTime`** — hour rows anchored at 08:00 while chips positioned from 08:30, so a booking sat half an hour from its own label. Both now share one anchor.
- **Short bookings** were packed by their real end but drawn at a 20-minute floor, so the next booking could overlap them on screen. Packing uses the rendered span.
- **Duplicate resource ids** now resolve last-wins, matching `Scheduler`'s own lookup so the two components cannot disagree about a colour.

**On the ARIA finding I went further than the suggestion.** You offered either restructuring to hour-rows or using semantics that match the column-oriented DOM. I took the second, and for a reason worth stating: a `grid` role also promises two-dimensional arrow navigation, which this read-only view does not implement — so it was over-claiming twice, not once. Each column is now a labelled list of bookings, which is what it actually is.

**Layout** — columns keep an 8rem minimum and the header shares a single horizontal scroller with the body, so a long resource list scrolls instead of compressing; two scrollers would have drifted and put every column under the wrong heading.

Eight new tests, including one per crash. Clean Release build 0/0, full non-E2E suite green, both CSS bundles rebuilt, tree scanned for leftover conflict markers.

Brain2k-0005:
@codex review

PR #414, opened 2026-08-13
feat(scheduler): make the first-party views reachable from <Scheduler>

Brain2k-0005:
The first-party views existed, were tested, and were demoed on the docs page — but the **shipped `<Scheduler>` could not render them**. A consumer using the public component therefore had no way to reach week numbers, weekend hiding, live-region announcements, resource columns, or `SchedulerEvent.Recurrence`, which the FullCalendar wrapper ignores outright (its `ToJsEvent` branches solely on the legacy `DaysOfWeek` pair, so a structured rule reaches change detection and nothing else).

## Why an opt-in and not a switch of the default

Flipping the default would change what **every existing consumer renders on an upgrade** — different markup, different behaviour, a different engine. That is a product decision for the repo owner, not a cleanup, and I have deliberately not taken it.

What an opt-in *can* do without deciding anything: make the capability reachable. `Engine` defaults to `SchedulerEngine.FullCalendar`, and the first test asserts that with it unset nothing changes — no first-party markup, no live region. Everything else in this PR is behind an explicit choice.

## What `Engine=FirstParty` actually does

**It skips the JS init entirely** — no instance is created, no calendar library loads. That is what makes it a genuinely dependency-free path rather than the same wrapper with different markup, and a test asserts the init call count is zero. Toolbar navigation moves an internal anchor date instead of delegating to FullCalendar.

View mapping: Month → `SchedulerMonthView`, Week/Day → `SchedulerTimeGridView` with `Days=7/1`, List → `SchedulerAgendaView` (the same flat chronological reading `listWeek` gives).

## Stated gaps, up front

`SlotDuration` is not carried across, and the JS-side imperative navigation has no equivalent. Both are documented **on the `Engine` parameter itself**, so the gap is visible before someone switches rather than discovered afterwards.

## Verification

Clean Release build 0 errors / 0 warnings, full non-E2E suite green, docs-parity exit 0. Disable-check: neutering the engine branch fails the six behavioural tests and leaves the two "default must not change" tests green — which is the split that matters here.

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in first-party scheduler engine with month, week, day, agenda, and resource views.
  - Added week numbers, weekend visibility controls, resource navigation, event styling, and drop validation.
  - Preserved FullCalendar as the default scheduler engine.
- **Localization**
  - Added translated “Resources” labels across supported languages.
- **Documentation**
  - Added setup guidance and examples for first-party scheduler features.


Brain2k-0005:
@codex review

Brain2k-0005:
All ten fixed in `3df845e8`. One of them is a credibility problem with the PR itself, so let me lead with it.

**The `Engine` documentation advertised week numbers, weekend hiding and resource columns. None of the three was reachable.** No parameters existed for the first two, and no branch rendered `SchedulerResourceView` at all. I wrote the claim into the public API docs and did not implement it — the same shape as #410's announcement that never fired, except here it was a promise a consumer would read before switching. All three are now real: `ShowWeekNumbers`, `HideWeekends`, and a `SchedulerView.Resource` value (appended, so existing values keep their numeric identity) wired to the resource view, offered in the toolbar only where it works.

**P1 — edits reached nobody.** Forwarding `OnEventChange` straight to the children skipped `EventsChanged`, so a consumer using `@bind-Events` without a separate handler saw the chip snap back. Edits now route through the component's own state.

The rest, each a case where switching engines silently lost something the wrapper does:
- **Resource colour fallback** — passing `EventColor` raw dropped the `SchedulerResource.Color` step, so an event carrying only a `ResourceId` lost its colour on switching.
- **`EventClassNames` and `SchedulerEvent.ClassNames`** never reached the views.
- **Agenda window** started wherever the anchor sat instead of the configured week the toolbar title names.
- **Time-grid PageUp/PageDown** were not wired to the toolbar navigation.
- **The cached title** did not follow a UI-culture change — the wrapper re-reads its title from FullCalendar, this engine owns it.
- **The `Events` XML doc** had been detached by my parameter insertion, leaving `Events` undocumented and `Engine` with two summaries.

---

**On the runtime engine switch, my first fix broke the thing this PR promises.** I relaxed the `firstRender` gate for every FullCalendar render, which failed `SchedulerInitRaceTests.Culture_changed_while_init_is_in_flight_is_not_lost` on both frameworks — a deterministic break of the default path, in a PR whose central claim is that the default path is untouched. The extra init pass is now scoped to an actual `FirstParty → FullCalendar` transition and nothing else.

Four new tests cover the advertised-but-absent features and the bind-Events routing. Clean Release build 0/0, full non-E2E suite green, docs-parity exit 0.

Brain2k-0005:
@codex review

Brain2k-0005:
All six fixed in `a32b45f3`. The packaging one was the most consequential thing found in this PR, and it was a consequence of the PR's own premise.

**Wiring the first-party views into `Scheduler.razor` broke `lumeo add scheduler` for everyone.** The shipped component now referenced `SchedulerViews/*` and `Kernel/*`, both of which `RegistryGen` deliberately excludes — so a source-vendored install would have produced code that does not compile, and not only for opt-in users. This is the same trap as the RRULE parser two PRs ago, except that one affected a single new file and this one affected the whole component.

Fixed at the root rather than per-file: the views and the kernel moved into `UI/Scheduler/`, which is where the generator looks, and the `SchedulerViews` exclusion is gone. Verified by regenerating and asserting the manifest contains all seven files the wrapper needs — 15 files total, none missing.

**P1 — and a correction to my own previous fix.** I "fixed" edit routing last round by building a local list and pushing it through `EventsChanged`. That repaired the controlled case and left the **documented uncontrolled mode exactly as broken as before**: with no delegate, the edit went into a temporary list and was dropped. The working copy is updated first now, and the views read that copy rather than the raw parameter. Two rounds, two P1s, same method — worth naming.

The rest:
- **Switching an initialized wrapper to FirstParty** left the JS calendar alive behind the new markup, holding its `DotNetObjectReference`. It is destroyed now.
- **Resource navigation** moved seven days per click on a view that renders one date.
- **`SchedulerView.Resource` under the wrapper** sent an unknown view name to FullCalendar instead of the Day fallback its own documentation promised.
- **The docs page** still said there was no switch to opt into.

Clean Release build 0/0, full non-E2E suite green, docs-parity exit 0.

Brain2k-0005:
@codex review

Brain2k-0005:
All eight fixed in `31d9293e` — the P1 first: the first-party engine never sets `_initialized`, so the event-adoption block skipped it and a parent replacing its `Events` collection saw the calendar keep rendering the old one **forever**. Also: the `DotNetObjectReference` is disposed with the JS instance, a FullCalendar init error no longer blocks an engine that needs no JS, the Resource branch requires a non-empty list, `CanDrop` is exposed and forwarded, the agenda view gained the class resolver its siblings already had, and List/Resource titles describe the window each actually shows.

---

**I want to be straight about this PR's state rather than just reporting another green run.**

Three review rounds: 10, then 6, then 8 findings — **a P1 in every one**. And the pattern matters more than the count:

- Round 1: I advertised week numbers, weekend hiding and resource columns in the API docs and shipped none of them.
- Round 2: my fix for edit routing repaired the controlled case and left the documented uncontrolled mode exactly as broken.
- Round 3: the adoption path was missing for the same reason the earlier ones were — I kept fixing the *case named in the finding* instead of the *mechanism*, and the next layer surfaced each time.

It also grew far past its title. What began as "add an opt-in parameter" now relocates eleven files, adds a view value, four parameters and a resolver, and touches three components — because wiring the views into `Scheduler.razor` made the shipped component depend on files `RegistryGen` excludes, which would have broken `lumeo add scheduler` for everyone.

Everything above is fixed and verified — clean Release build 0/0, full non-E2E suite green, docs-parity 0, registry regenerated and asserted to contain all fifteen required files. But on this evidence I would not merge it on a green CI alone. **My recommendation is that this one waits for a review round that comes back clean**, unlike the five ReUI PRs already on master, each of which converged. If you would rather split it — the file relocation as its own PR, the opt-in on top — say so and I will restructure it that way.

Brain2k-0005:
@codex review

Brain2k-0005:
Round 4's five findings are fixed in `3ffbb7d3`, and `e81b937f` closes a gap the review round did not raise but the PR had: `Engine` was documented in prose with no live demo, so the integration this PR exists for had no surface a developer could compare.

The P1 — the pushed-events hash was recorded *after* the parent callbacks ran, so a controlled parent that rejected an edit by handing its old collection back looked like an unrelated change and had its rejection adopted as the new truth. It now follows the same order the wrapper's own commit path already used.

The remaining four: the anchor tracks wrapper navigation (so switching engines keeps the date you are looking at); first-party Today resolves the **browser** date, since a Blazor Server circuit can sit on a different calendar day; the resource view gained the class resolver its sibling views have; and `Height` bounds the scroll viewport instead of only capping the outer box.

The new demos were verified headless against the running docs site rather than by inspection: month renders 42 cells with 6 week-number rowheaders, the toolbar's Week button empties those and swaps in time-grid markup, Next re-renders, and the resource board lays out 3 lanes — no page exceptions on any step.

Brain2k-0005:
@codex review

Brain2k-0005:
Rebased onto master now that #415 has landed, so this PR is down to what it is actually about: the wrapper integration. The only real-source conflict was a RegistryGen comment both sides worded the same way; the rest were generated artifacts, which were reset to master and rebuilt rather than hand-merged.

@codex review

Brain2k-0005:
Confirmed and fixed in `880e537a`. It is a consequence of my own earlier P1 fix: opening adoption to the first-party engine — without which a parent replacing its `Events` collection was never picked up — also let that path reach an interop call guarded by nothing but a null-forgiving `!`. The claim was untrue there, since this engine never initializes FullCalendar.

The JS side no-ops on an unknown id, which is exactly why it looked harmless. Getting there still imports `scheduler.js`, and where JS is unavailable at all that import can throw and take an otherwise pure-Blazor update down with it.

The test asserts both halves, because either one alone would pass a broken implementation: the replacement **is** adopted, and the JS bridge is never touched. Reverting the guard fails it.

@codex review

Brain2k-0005:
Both addressed:

- **CI (registry stale)** — the registry vendors component source verbatim, so editing a shipped `.razor` invalidates it. I had not regenerated after the guard went in. Fixed in `4228ee44`.
- **CodeRabbit's assertion point** — correct, and the same trap I walked into elsewhere in this stack. `Assert.Contains("Standup", cut.Markup)` passes when the chip renders anywhere on the page, and the title also appears in its tooltip and aria-label, so the resource binding this test exists to prove was the one thing it never checked. Now scoped to the column's own text.

Codex came back clean on `880e537a`.

@codex review

PR #415, opened 2026-08-13
fix(scheduler): ship the first-party views with lumeo add scheduler

Brain2k-0005:
**Split out of #414**, which had grown from "add an opt-in parameter" into a cross-cutting change touching three components. This is the structural half, and it stands on its own.

## The gap

`SchedulerMonthView`, `SchedulerTimeGridView`, `SchedulerAgendaView`, `SchedulerResourceView` and the kernel they run on are **public API** — documented, demoed on the docs page, independently usable. A source-vendored install received **none** of them: `RegistryGen` builds a component's `files` list from its own `UI/<Name>/` folder, and these lived in `UI/SchedulerViews/` and `Kernel/`.

Same trap the RRULE parser hit in #409, one directory wider. It also blocks #414 outright: the moment `Scheduler.razor` renders these views, `lumeo add scheduler` would produce source that does not compile — for **every** consumer, not just opt-in ones.

## What this changes

A pure relocation into `UI/Scheduler/`, plus removal of the now-pointless `SchedulerViews` exclusion in the generator. That exclusion existed because the views were work-in-progress with no docs page; neither has been true since they were documented.

**No namespace, behaviour or API change** — the diff is 12 renames and 11 deleted lines in the generator.

## Verification

Manifest regenerated and asserted to contain all nine view/kernel files (15 total, none missing). Clean Release build 0 errors / 0 warnings, full non-E2E suite green, docs-parity exit 0.

## Why separately

#414's four review rounds produced 29 findings, every one of them in the *integration* (`Scheduler.razor`) and **none** in this relocation. Landing the structural half now shrinks that PR to the part that genuinely needs more review.


Brain2k-0005:
@codex review

Brain2k-0005:
Fixed. Good catch — and it is a consequence of the relocation itself, which makes it exactly the kind of thing splitting this PR was meant to surface.

Moving `.cs` files into a scanned `UI/` folder fed their source text to the tag scanner, which treats any `<Word` as a component reference. `Array.Empty<List<Block>>()` contains the literal text `<List`, so the manifest declared a dependency on the `list` component; `popover` came from a `// Not <Popover>` comment of mine.

**Generics are fixed at the scanner**, precisely: a real Razor tag never has an identifier character immediately before its `<`, and a generic always does, so a negative lookbehind separates them cleanly. I verified this did not silently drop real dependencies elsewhere — comparing every component's dependency set against `master` shows **zero edges gained and zero lost**, so the only removals are the two the relocation would have introduced.

**The comment is reworded rather than the scanner loosened.** Line comments are deliberately not stripped (the code says why: a false negative loses a real dependency, which is worse), so naming a tag in one still counts. The comment now says "Popover/PopoverTrigger" without brackets, with a note explaining the constraint so the next person does not reintroduce it.

Manifest: `['button', 'toggle-group', 'tooltip']` — the three the component genuinely uses. Clean Release build 0/0, full non-E2E suite green, docs-parity 0.

Brain2k-0005:
@codex review

PR #416, opened 2026-08-13
feat(scheduler): make the first-party grid interval configurable

Brain2k-0005:
Stacked on #414 — base is that branch, not master, because it touches the same first-party time grid. Rebase to master once #414 lands.

## Why

ReUI's event calendar lists `grid interval and snap duration` under its display controls. Lumeo's first-party time grid was fixed at one hour per row and snapped at a hardcoded 15 minutes — and the docs admitted it, listing `SlotDuration` as ignored on this engine. An acknowledged gap in shipped documentation is the kind a developer comparing the two finds immediately.

## What

`SlotDuration` is honoured, with `SnapDuration` alongside it for drags that should be finer than the grid.

The row *is* the cell, so this is not a cosmetic set of guide lines — keyboard navigation, aria-labels, click targets and drag-create all move at the configured interval. Hour labels stay in their own gutter, one per hour, and hour boundaries keep a stronger rule than the lines inside them, so a 15-minute grid still reads as hours rather than 96 identical stripes.

Rows have to tile an hour exactly or the gutter drifts away from the lines it names. A value that does not divide 60 is rounded up to one that does, and a shifted `SlotMinTime` still starts the grid on a whole hour.

## Two bugs this had to fix to work at all

- The click handler passed the clock **hour** where a **row index** was expected, so with `SlotMinTime` at 08:00 the roving tabindex landed eight rows below the cell actually clicked. Pre-existing; sub-hour rows just made it visible.
- Snap was missing from the drag-registration hash, so changing the interval after first render redrew the grid while JS kept snapping to the old value.

## Defaults are untouched

With `SlotDuration` unset: the same 24 rows, the same 15-minute snap, the same 30-minute range from a double-click. Three of the new tests exist only to pin that, and they are the three that still pass when the feature is disabled — the other 13 fail, which is how I know they test something.

## One finding worth repeating

The first attempt used `border-border/15`. Tailwind only generates opacity utilities for literals it has seen, and that one is not in the bundle — so `border-b` fell back to its default colour and the sub-hour lines rendered **darker** than the hour rules. The class-name assertion passed the whole time; only the browser showed it. The view now uses a generated utility, and a test reads the compiled bundle to keep it that way.

Live-verified: 96 rows of 12px, gutter still 24 labels, computed border alpha 0.3 on the hour against 0.2 inside it, no page exceptions.

@codex review


Brain2k-0005:
Merged #414's P2 fix up the stack (`a3237d54`); no conflicts.

@codex review

Brain2k-0005:
Rebased onto master now that #414 has landed, and the base is switched to `master`.

Every conflict was the same shape: master carries #414 squashed, this branch carries the identical content as the commits it was reviewed in, so git could not match them. HEAD is #414 plus `SlotDuration` in each case — checked hunk by hunk rather than taken on faith. Generated artifacts were rebuilt with RegistryGen rather than hand-merged.

@codex review

Brain2k-0005:
Superseded by #418 — same branch, same content, retargeted to `master`. This PR was auto-closed when its base branch was deleted on merging #414; passing `--delete-branch` on the base of a stacked PR was my mistake, and #417 will be retargeted before the next merge rather than after.

Issue #413, opened 2026-08-13
Flaky: GanttV3KeyboardNavigationTests.Mid_Flight_Combination_Drift under parallel load


Brain2k-0005:
`GanttV3KeyboardNavigationTests.Mid_Flight_Combination_Drift_Reregisters_With_Live_Rules_Instead_Of_Unregistering` fails intermittently under load and blocks unrelated PRs.

**Evidence gathered 2026-08-13**
- Failed on #406, #409, #410 (twice), #411 (locally) and #412 — **none of which touch Gantt code**.
- Passes in isolation and at class level, every time.
- Passes on unmodified `master`, locally (3 runs) and in master's own CI.
- Only fails in the full-suite run at maximum parallelism, and typically right after a clean rebuild — i.e. peak CPU contention.
- Two sibling tests show the same shape occasionally: `ToastStackingTests.Focus_Moving_Between_Two_Controls_In_The_Group_Never_Collapses_It` and `TooltipClickDismissAndArrowAnchorTests.After_A_Mouse_Click_The_Tooltip_Stays_Closed_Until_The_Pointer_ReEnters`.

**What did NOT fix it**
Raising the spin-wait budget from 1s to 6s (`for (i < 600) await Task.Delay(10)`, merged in #410). It still fails, which rules out "the second registration is merely slow" — under contention it appears not to happen at all within the window. That is a different cause from the one the budget addressed, so the diagnosis behind that change was incomplete.

**Why it matters beyond the noise**
On #406 this same test failure pointed at a *real* defect: an interop resolve guarded on outcome rather than attempt, running on every render (7 calls instead of 1). A flake that sometimes signals a genuine bug is worse than a reliably failing one — it trains everyone to re-run.

**Suggested direction**
Replace the spin-wait with bUnit's `WaitForAssertion`, or make the drift correction observable through a completion signal rather than a polled count, so the test does not depend on scheduler timing at all.

Brain2k-0005:
Fixed in 5.10.0 (#477). The mechanism was an overlapping-render race in `GanttBar`'s prevent-default-keys registration: a second render dispatching while the first registration was mid-await re-checked stale state and issued a redundant register call, visible only under thread-pool contention. Each registration now stamps a generation before its first await and a superseded one bails out, so exactly one performs the correction regardless of completion order. A deterministic two-gate test reproduces the interleaving.

PR #417, opened 2026-08-13
feat(scheduler): add the N-day view and a display time zone

Brain2k-0005:
Stacked on #416 (which is stacked on #414). Closes the last two ReUI capabilities the first-party engine was missing, so 5.0.0 ships the whole set.

## N-day — `SchedulerView.MultiDay` + `VisibleDays`

A rolling window, not a week. It starts at the anchor instead of a week boundary, and prev/next page by the window's own width so consecutive pages tile the calendar rather than overlap.

Seven days is the awkward case: that is exactly the width at which the time grid snapped to the configured week start, so a rolling seven-day window silently became "this week". The grid gained `AlignToWeek`, true everywhere it was before. The test for it was written before the fix and failed for exactly that reason.

The toolbar button appears only once a width is chosen — an N-day view with no N is not a view — and it is labelled with the actual count.

## Time zone — `TimeZone`

The hard part was not converting; it was deciding what a `DateTime` in a `SchedulerEvent` means. Every view is deliberately a wall-clock renderer that never touches `TimeZoneInfo`, which is what makes the grid immune to DST and to the gap between a CI box and a dev machine. That property is kept: the views stay pure, and the projection happens at the wrapper's edges.

`DateTimeKind` therefore decides what can move:

- `Utc` / `Local` — real instants, re-read in the display zone.
- `Unspecified` — a wall-clock reading already in whatever zone the caller meant. **Left alone.** This is what virtually every existing caller passes, so turning a zone on cannot silently shift their data.

All-day events are dates, not instants, and never move — shifting one by hours moves it onto a different day.

Edits are converted back *before* they reach `OnEventChange`/`EventsChanged`, using the **original** event to recover the frame, because the views hand back naive values and cannot know it. Getting that backwards would rewrite the caller's data into a different frame on every drag.

Degrading rather than failing: an unknown zone id renders the caller's own values instead of throwing, and a drag landing in the hour a spring-forward transition skips is nudged past the gap rather than surfacing as an unhandled exception on the circuit.

The now-indicator is computed in the browser from the real clock, so it takes the zone too — via `Intl`, which is the only IANA database a browser has. Only IANA ids travel; `Intl` does not understand Windows ids and would silently leave the line on the viewer's own clock.

## Also fixed

Found while wiring the title: the **initial** title came from the wrapper-shaped formatter, which has no branch for `List`, `Resource` or `MultiDay`. All three opened on the month name and only corrected themselves once the user pressed prev or next. Pre-existing; MultiDay is what exposed it.

## Verification

Against the running docs site, not by inspection: the same UTC events render **09:00/13:00 in Europe/Berlin and 16:00/20:00 in Asia/Tokyo** (07:00 and 11:00 UTC in August), the N-day grid shows three columns from today with its own toolbar entry, and there are no page exceptions on either.

The one suite failure in the full run was `Mid_Flight_Combination_Drift` — the known load-sensitive flake tracked in #413, in the Gantt keyboard tests this branch does not touch. It passes 3/3 in isolation.

@codex review


Brain2k-0005:
Merged the stack down to here (`994410a0`); no conflicts.

@codex review

Brain2k-0005:
Retargeted to `master` and brought up to the rebased grid-interval branch. Only generated artifacts conflicted; they were rebuilt with RegistryGen rather than hand-merged, and I verified afterwards that all three layers survive — #414's interop guard, the grid-interval rows, and this PR's MultiDay view and display time zone.

Note this PR now sits on top of #418 (which replaced #416 after GitHub auto-closed it — see there), so it should land after that one.

@codex review

Brain2k-0005:

@codex review

PR #418, opened 2026-08-13
feat(scheduler): make the first-party grid interval configurable

Brain2k-0005:
Replaces #416, which GitHub auto-closed when its base branch was deleted on merging #414 — my mistake for passing `--delete-branch` on the base of a stacked PR. Same branch, same content, now targeting `master`. #416 carries the review history: Codex came back clean on `6417f71a` and again on `a3237d54`.

## Why

ReUI's event calendar lists `grid interval and snap duration` under its display controls. Lumeo's first-party time grid was fixed at one hour per row and snapped at a hardcoded 15 minutes — and the docs admitted it, listing `SlotDuration` as ignored on this engine. An acknowledged gap in shipped documentation is the kind a developer comparing the two finds immediately.

## What

`SlotDuration` is honoured, with `SnapDuration` alongside it for drags that should be finer than the grid.

The row *is* the cell, so this is not a cosmetic set of guide lines — keyboard navigation, aria-labels, click targets and drag-create all move at the configured interval. Hour labels stay in their own gutter, one per hour, and hour boundaries keep a stronger rule than the lines inside them, so a 15-minute grid still reads as hours rather than 96 identical stripes.

Rows have to tile an hour exactly or the gutter drifts away from the lines it names. A value that does not divide 60 is rounded up to one that does, and a shifted `SlotMinTime` still starts the grid on a whole hour.

## Two bugs this had to fix to work at all

- The click handler passed the clock **hour** where a **row index** was expected, so with `SlotMinTime` at 08:00 the roving tabindex landed eight rows below the cell actually clicked. Pre-existing; sub-hour rows just made it visible.
- Snap was missing from the drag-registration hash, so changing the interval after first render redrew the grid while JS kept snapping to the old value.

## Defaults are untouched

With `SlotDuration` unset: the same 24 rows, the same 15-minute snap, the same 30-minute range from a double-click. Three of the new tests exist only to pin that, and they are the three that still pass when the feature is disabled — the other 13 fail, which is how I know they test something.

## One finding worth repeating

The first attempt used `border-border/15`. Tailwind only generates opacity utilities for literals it has seen, and that one is not in the bundle — so `border-b` fell back to its default colour and the sub-hour lines rendered **darker** than the hour rules. The class-name assertion passed the whole time; only the browser showed it. The view now uses a generated utility, and a test reads the compiled bundle to keep it that way.

Live-verified: 96 rows of 12px, gutter still 24 labels, computed border alpha 0.3 on the hour against 0.2 inside it, no page exceptions.

@codex review

## Summary by CodeRabbit

- **New Features**
  - Added configurable scheduler slot durations for day, week, and resource views.
  - Added configurable snapping intervals for dragging, resizing, and event creation.
  - Improved time-grid support for custom intervals, including labels, styling, accessibility, keyboard navigation, and selection.
  - Added a first-party Scheduler demonstration showcasing 15-minute slots and drag snapping.

- **Bug Fixes**
  - Improved event creation accuracy for sub-hour time slots.

- **Documentation**
  - Updated Scheduler documentation to describe supported slot and snap duration settings.


Brain2k-0005:

@codex review

PR #419, opened 2026-08-13
feat(scheduler)!: remove FullCalendar

Brain2k-0005:
Stacked on #417. **Breaking**, and deliberately aimed at 5.0.0.

`Scheduler` renders its own Blazor views and nothing else. Gone: the wrapper, its `Engine` switch, `scheduler.js`, ten interop methods, five CDN dependencies, and the vendored copies of them.

## What breaks

- **`Engine` and `SchedulerEngine`** — there is one renderer, so there is nothing to choose.
- **`OnInitError`** — there is no init to fail.
- **`JsOnEventClick` / `JsOnDateSelect` / `JsOnEventChange`** — the bridge that invoked them is gone.
- **`Scheduler`'s `IAsyncDisposable`** — it existed only to destroy the JS instance.

Everything else maps: `InitialView`, `InitialDate`, `Height`, `@bind-Events`, `OnDateSelect`, `OnEventChange`, `BusinessHours`, `NowIndicator`, `SlotMinTime`/`SlotMaxTime`/`SlotDuration`, `Resources`, `EventColor`, `CanDrop`. What used to require the opt-in engine — recurrence, week numbers, weekend hiding, live announcements, the resource view, the N-day view, the display time zone — is simply what the component does.

## Two real bugs this surfaced

Both in the path that survives, both found by porting the wrapper's tests rather than deleting them:

- **An ambiguity guard lived only in the deleted bridge.** A blank or duplicate event Id merged a drag onto the first matching record. The commit path that replaced it used the same bare `FindIndex` the original finding (#21) was about — so removing the wrapper would have quietly dropped the fix while leaving the defect. The guard came across with its tests.
- **The commit path corrupted its own change discriminator.** It wrote the locally-edited hash into the field that tracks what the `Events` *parameter* last was, so in uncontrolled mode the parent's very next re-render — re-supplying the collection it always passes, for some unrelated reason — looked like a genuine change and clawed the user's drag back. The bridge had this right; the first-party path did not.

Plus one inconsistency: the toolbar title formatted against `CurrentCulture` while every view and the change detection use `CurrentUICulture`, so a host that sets them apart got German buttons over an English month name.

## Tests

Ported wherever they described behaviour that still exists — controlled/uncontrolled rollback, the change-detection hash's field coverage, toolbar navigation, the culture-change title refresh. They now assert on **what the grid renders** instead of counting interop calls, which is a stronger check than the originals were: the rollback test proves the chip is back on the bound day, not that some method was called.

Deleted only what described the bridge itself: the init-handshake races, the module import path, the init options, and the FullCalendar locale-code mapping (including the "iv" trap, which was about a locale pack nothing fetches now).

## Verification

In the browser, not by inspection: the docs page issues **no request matching `/fullcalendar/i`**, its text no longer mentions it, and all nine grids render with no page exceptions. Full gate clean: 0 errors, suite green.

@codex review


Brain2k-0005:

@codex review

Brain2k-0005:
Auto-closed one second after #417 merged: its base branch was deleted by that merge. Continued as #420 against `master` — same branch, same content. My mistake, and the second time in this stack: I flagged it after #416 and then retargeted only #417, not this one.

PR #422, opened 2026-08-13
fix(scheduler): three reported view defects

Brain2k-0005:
Everything you reported except the multi-day spanning bar, which is a month-layout rewrite and gets its own PR.

## 1. Day view collapsed to 82px

All four views' roots carry no width basis, so inside a flex parent they become a flex item with `flex: 0 1 auto` and shrink to content. Seven day columns hid it; one did not.

Measured rather than guessed — the parent chain on the live page:

| element | width | display |
|---|---|---|
| demo preview container | 926px | flex |
| view root | **82px** | flex item |

Fixed with `w-full` on **all four** roots, not just the time grid — none of them should depend on how a host lays them out.

## 2. Now-indicator only spanned one day

The line was appended **into** each day column, so it could only ever be that column wide, and a seven-day grid grew seven stubs. It is now **one** line on the grid host, handed every visible date so it can decide whether today is on screen at all.

That also fixes the crash you pasted. The per-column loop indexed `_dayColumnRefs`, an array rebuilt on every parameter set; an entry not yet bound reached JS as an object with no `appendChild` — `containerEl.appendChild is not a function`, an unhandled render exception. One registration on the host has no unbound-index case. The JS guard is hardened as well, because it runs inside an interval that would otherwise never be cleaned up.

## 3. "Danach passiert nichts"

Drag-create drew a ghost in views where nothing listens; releasing it did nothing, because `CommitCreate` reaches a callback with no delegate and drops the range. Now gated on `OnDateSelect` having one — an explicit `Selectable="false"` still wins.

I checked whether this was my doing: the week demo **never** had an `OnDateSelect`, before or after the FullCalendar removal. Not a regression, a component offering a gesture that cannot work.

## A note on how #3 is verified

My browser check proved nothing — a synthetic drag produced zero ghosts **with and without** the fix. I only caught that by running the same script against live master as a control. It is asserted on the options pushed to JS instead, where the change actually lives. Reverting the gate fails exactly the two tests that should fail and leaves the other three green.

The existing `NowIndicator_True_Registers_One_Line_Per_Day_Column` test pinned the behaviour you reported as wrong, so it is rewritten to the new contract rather than deleted.

## Verification

Live: day-view grid **804px** (was 24), now-line **804px** anchored in the grid, **one** instead of seven, no page exceptions. Full gate: 0 errors, suite green.

@codex review

## Summary by CodeRabbit

- **New Features**
  - Scheduler agenda, month, resource, and time-grid views now use the full available width.
  - Drag-to-create date selection is enabled only when selection is supported and a date-selection handler is configured.
  - The current-time indicator now works consistently across multi-day views.

- **Bug Fixes**
  - Improved scheduler interaction handling and prevented unsupported drag gestures.

- **Tests**
  - Expanded Scheduler coverage with additional tests for date-selection behavior and time indicators.


Brain2k-0005:
@codex review

PR #423, opened 2026-08-13
fix(scheduler): draw a multi-day event as one bar across its days

Brain2k-0005:
Your fourth report — the two-day event that showed a titled chip on day 5 and an empty bar on day 6.

## What was actually wrong

The lane model was already right: it marks continuation segments and drops their rounded corners on the joining sides precisely so a run reads as one bar. They just never touched. A day cell pads its content by `p-1` (4px) and carries a 1px right border, so consecutive segments of the same run sat **9px** apart — and since only the run's first day carries the title, the rest looked like empty stubs.

Continued segments now bleed out to the cell's border box on the side they continue.

| | gap between segments |
|---|---|
| before | 9px |
| after | **0px** |

Measured on the rendered page, not reasoned about.

## The part I got wrong first

My first attempt closed only 4px of the 9. The pill is `w-full` — `width: 100%` — and a negative margin **shifts** that box without widening it. The width has to grow by the same amount it bleeds. I only caught it because I measured again instead of trusting the change.

## Deliberately not joined

Separate occurrences of a **recurring** event. Those are distinct instances that happen to fall on adjacent days, not one run, and merging them would misrepresent the data. A test pins that, and a single-day event keeps its exact cell width.

## Verification

Visual: the "Hackathon" event now renders as one continuous bar across the 17th–18th with a single title. Disable-check: removing the bleed fails exactly the three tests that assert it and leaves the three behavioural ones green. Full gate: 0 errors, suite green.

Stacked conceptually on #422 (the other three reports) but independent — both target `master`.

@codex review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multi-day scheduler events so adjacent segments connect seamlessly without visual gaps.
  * Preserved correct styling for single-day events and separate recurring occurrences.
  * Ensured event titles appear only on the first segment of a multi-day event.
  * Improved visual continuity across event segments while maintaining existing color styling.

* **Tests**
  * Added coverage for multi-day event rendering and segment styling.


Brain2k-0005:
@codex review

Brain2k-0005:
Rebased onto master now that #422 has landed. Both touch `SchedulerMonthView`, but the source merged cleanly — only the generated registry artifacts conflicted, and those were rebuilt rather than hand-merged. Verified both sides survive afterwards.

@codex review

PR #424, opened 2026-08-17
feat(scheduler): add a resource timeline

Brain2k-0005:
Your **3** — rows are resources, the time axis runs **horizontally** across days, weeks or months. The shape Outlook's scheduling assistant has, and the one Lumeo had no equivalent for.

## Why it is its own view, not an option on the existing one

Three layouts get called "resources side by side", and they answer different questions:

| | shape | answers |
|---|---|---|
| `SchedulerResourceView` | one day, a **column** per resource, vertical clock | who is where **today** |
| **`SchedulerTimelineView`** (new) | a **row** per resource, **horizontal** axis over days/weeks/months | how is this resource booked over a **stretch** |
| calendar panes (still to come) | several full week grids side by side | overlay vs side-by-side calendars |

Stretching the resource view to a week would mean a vertical axis seven days long — unusable. That is why this is a separate view rather than a parameter.

## On not depending on the Gantt

`GanttScale` already solves date→pixel for a horizontal axis, with a well-documented algorithm. It is **internal** to the Gantt package, and making `Lumeo.Scheduler` depend on `Lumeo.Gantt` would put a whole Gantt into every project that only wanted a calendar — the opposite of why FullCalendar was removed. The subset needed here is small, so the Scheduler carries its own.

## What the tests are for

Each of these is quietly wrong-*looking* rather than obviously broken, which is why they are pinned:

- **Overlapping bookings get their own lane** and the row grows. Drawing them over each other would hide exactly the conflict this view exists to surface.
- **A booking running in from before the window is clamped, not dropped.** A resource that is busy must never read as free.
- **The gutter's row heights are driven by the same value as the timeline's** — the two halves are separate elements, and if they drift the names stop lining up with the bars.
- **Month columns take their day fraction against the real month length.** A flat 30 drifts by a day inside February.

Disable-check: breaking lane packing and clamping fails exactly the three tests that assert them, and nothing else.

## A bug this turned up

`Resource` and `Timeline` both document that they degrade to the day view when `Resources` is empty. They did not — they fell through to the switch's default branch, which is the **agenda**. Caught by a test written against the documentation rather than against the code. Now an explicit case.

## Docs

The demo uses multi-day bookings. My first version reused the hour-long meetings the other demos share, and on a day-per-column axis those render as four-pixel slivers — correct, and useless for showing what the view does. Live-verified: 3 rows, 10 columns, 6 bars, one today marker, Alice's row 52px against the others' 30px because of her double booking, gutter heights matching exactly, no page exceptions.

## Still not in this PR

Drag and resize on the timeline. Reachable via `<Scheduler Engine…>`'s toolbar, `TimelineUnit` and `TimelineColumns`; bars are clickable. Say if drag matters before the calendar-panes work.

@codex review

## Summary by CodeRabbit

- **New Features**
  - Added a Scheduler timeline view with day, week, and month scales.
  - Supports resource rows, overlapping bookings, recurrence expansion, navigation, today indicators, event clicks, customizable sizing, and colors.
  - Added accessible labels and RTL-compatible positioning.
  - Scheduler falls back to a day grid when resources are unavailable.
  - Added localized “Timeline” labels across supported languages.

- **Documentation**
  - Added an interactive timeline example and embeddable usage snippet.

- **Bug Fixes**
  - Improved recurrence handling for long date ranges and date-boundary safety.
  - Improved detection of component references in complex code and markup.


Brain2k-0005:
@codex review

Brain2k-0005:
## Review round: all 14 findings closed

Component and wrapper fixes in `22f1f86`, one test per finding in
`SchedulerTimelineReviewTests` so each is pinned rather than merely applied. The
three that were real defects rather than hardening: recurrences were not expanded
(a twice-weekly booking rendered once and the resource read as free), lanes were
packed by the clock while bars have a minimum width (two sequential hour bookings
covered each other), and the track positioned from the physical left while its flex
header reversed under RTL.

## The registry finding needed a wider fix than the entry

`timeline.json` listed a Scheduler test because the matcher strips comments before
scanning for a type reference, but **not string literals** — and the test asserts a
toolbar button reads `== "Timeline"`.

Fixing only that entry would have left the rule intact, so `31d7e3d` blanks literals
before the scan. That removes **157 links across 58 components**, all of the same
kind — sampled and confirmed before regenerating:

| link | actually |
|---|---|
| `Card` → 8 HoverCard tests | `"Card body"` fixture text |
| `Label` → 24 form tests | `AddAttribute(1, "Label", ...)` — a parameter *name* |
| `Progress` → a Gantt test | `{"Progress":20}` in an embedded JSON payload |
| `Accordion` → the contract test | a name in its exemption array; the test is reflection-driven and covers every component equally, so the subset that appeared was arbitrary |

Nothing was added, so no association changed owner. Interpolation holes are exempt —
`$"{typeof(Sheet).Name}"` is code, and blanking it would trade this false positive
for a false negative. Seven cases in the matcher's unit matrix, including the two
easy to get backwards: an escaped `{{brace}}` pair stays text, and code after a
literal containing an escaped quote is still scanned.

Gate: clean rebuild 0 errors / 0 warnings, 7949 tests green on both TFMs, registry
regenerated last.

@codex review

Brain2k-0005:
## Round 2: six findings fixed, two declined with reasons

**Timeline view** — duplicate event ids no longer collapse. Ids are caller-supplied and
need not be unique (the resource view already treats duplicates as legitimate input), so
expansion now runs per source event instead of joining instances back through a
dictionary keyed on `Id`, which kept only the last of a colliding pair. `dir="ltr"` moved
onto the horizontal scroller as well: pinning only the track left the overflow container
with an RTL scroll origin, so a wide timeline opened at the far end of the range.

**Wrapper** — the title's invalidation state gained the effective first weekday and the
effective view; the toolbar's selected value now comes from `EffectiveView`, so the Day
fallback actually reads as pressed instead of leaving every visible item unpressed.

## The occurrence cap was the interesting one

The flat 500 counted candidates from the series' **DTSTART**, not from the render window,
so chunking the expansion could not have helped: a later chunk still walks the same
candidates to reach it. Any window longer than the cap was truncated — a 24-month
timeline asks for 730 daily occurrences, and the last eight months of an axis it had
already drawn rendered as free on a resource booked every day.

The budget now scales with the requested window, floored at the old 500 and ceilinged at
20 000, so a `COUNT`-less, `UNTIL`-less rule against an absurd range still stops.

That flat cap turned out to be what kept the candidate generators inside `DateTime`'s
range. With the budget able to run further, each advance guards its own overflow — and the
new test found a real crash the old ceiling had been hiding: a weekly series ending in the
last representable week threw on the **time-of-day** addition, not on the week step.

## Registry matcher: a scanner, not more regexes

Both reports were the same defect. A pipeline of stripping passes cannot express
"whichever construct opens first wins", so `var sep = "//"; Render<Lumeo.Sheet>();` lost
its reference to the comment pass, and `$"""...{typeof(Lumeo.Sheet)}..."""` lost its hole
to the raw-string pass. One left-to-right scanner has no ordering to get wrong. It also
learned Razor `@* *@` comments — that is where `gantt` ← `ScrollspyOneWayHost.razor` came
from, a "(state-on-data-change, Gantt-class)" note — and char literals, so `'"'` no longer
opens a string.

## Declined

- **`heading` ← `GoalBatchTests.cs`** is real coverage: the file calls
  `ctx.Render<Lumeo.Heading>(...)` twice.
- **`testCoverage.files` vs `tests`** are not meant to match. `ComputeTestCoverage` counts
  a component's **own folder**; `tests[]` also credits related files that render it. Gantt
  reads 88 vs 11 for the same reason, long before this PR.

Gate: clean rebuild 0 errors / 0 CS warnings, 7 958 tests green on both TFMs (up from
7 949), registry and skill catalog regenerated last.

@codex review

Brain2k-0005:
## Round 3: six fixed, two of them regressions from round 2

Raising the recurrence cap moved the failure mode rather than removing it, and you caught
both halves of that:

- **Monthly BYDAY budget.** `MonthlyCandidates` resolves each ByDay *entry*, so first-through-
  fourth Monday is four occurrences a month. Counting distinct weekdays left the cap at its
  floor — the same falsely-free tail round 2 was meant to remove, just at a different rule
  shape.
- **The final-year bound.** Comparing whole years treated zero as the maximum offset for any
  series starting inside 9999, so January yielded alone.
- **The duration addition.** With the budget able to reach the last representable day,
  `candidateStart + duration` threw; it is clamped now. The old flat cap had stopped decades
  short of it.

**Timeline view:** `DateTime.MinValue` as `RangeStart` underflowed the one-tick lookback
before any row rendered.

**Matcher, both in the new scanner:** `EndOfHole` ignored comments, so a brace inside one
terminated the hole and blanked the expression after it. And an apostrophe in ordinary Razor
markup text was read as an unterminated char literal, blanking the rest of the line and the
component tag on it. Recognising Razor code regions properly means parsing Razor; the
literal's own shape — a single character or a backslash escape, closed before the line ends —
is the cheaper discriminator, and a genuine `'"'` still parses. Pinned in both directions.

Also on this push: the aggregate manifests were stale while the per-component file was
current. `docs/wwwroot/registry.json` and `registry-search.json` come from MSBuild targets in
`Lumeo.Docs.csproj`, which is not in `Lumeo.slnx` — so they never refresh in a local build,
and only CI's hand-mirrored copy step compared them. My gate now runs CI's exact sequence
twice and requires the second pass to be a no-op.

Gate: 0 CS errors / warnings, 7 962 tests green on both TFMs.

@codex review

Brain2k-0005:
## Round 4: all four fixed

**Brace parity.** You are right that `run != 2` was the wrong rule — every even-length run is
escaped text all the way down, so `$"{{{{Sheet}}}}"` is literal. It reads parity now
(`run >= dollars && (run / dollars) % 2 == 1`), pinned in both directions: four braces stay
text, three still open a hole.

**Timeline endpoints.** `TotalWidth` and `AxisEnd` both advanced blind past the final column.
They go through one clamped `Advance` helper now — same class as the `DateTime.MinValue`
underflow from round 3, at the other end of the range.

**The boundary burst.** Fixed by reserving a whole `perWeek`/`perMonth` cluster instead of a
flat two.

Worth recording, since it changes what the test asserts: your example day, **2020-06-30, is
in an inactive week** for an every-other-week rule anchored on Monday 1990-01-01 — 1591 weeks
from the anchor. So it has no occurrence regardless of the budget. The last one actually
inside the range is **Sunday 2020-06-28**, and that is the one an under-budgeted cap drops.
The test asserts that day, and its disable-check confirms it: reverting the margin turns it
red.

**Orphan expansion.** The displayed resource-id set is built first and unmatched events are
excluded before the expander is called.

Gate: 0 CS errors / warnings, 7 965 tests green on both TFMs, registry regenerated last and
idempotent on a second pass. The earlier `e2e` red was
`GanttV3WheelZoomTests.CtrlWheel_Anchors_The_Zoom_On_The_Pointer` — unrelated to this PR, it
passed on re-run of the same commit, and I recorded the evidence on #390, which would explain
the shape.

@codex review

Brain2k-0005:
## Round 5: all four fixed

Three of these were the same incomplete fix of mine. Round 4 clamped the timeline's
*endpoints*, so the throw just moved one call inward:

- **`BuildColumns`** advanced with a raw `Add` — which fails *first* for any column count above
  one, i.e. exactly the larger-column case the endpoint fix was meant to cover.
- **`AlignOrigin`** steps **backward** to the week start, so with a first weekday other than the
  one the anchor falls on it goes below `DateTime.MinValue`.
- **`TimelineTitle`** formats the same axis and threw during `OnInitialized` while the view
  itself already rendered.

All four advances go through one clamped helper now, in both directions.

**Razor attribute expressions.** Correct, and the cause is that a markup attribute is delimited
by the same quote a C# string uses, so the scanner blanked the whole value. Distinguishing
markup from code properly means parsing Razor; keeping the `@`-expression itself — `@(...)`
with balanced parens, or an implicit `@Foo.Bar` chain, **in `.razor` files only** — is the
cheaper equivalent. Pinned in all four directions: explicit and implicit expressions count,
plain attribute text stays text, and the same string in a `.cs` file is unaffected.

Worth noting: regenerating the registry after this moves **no association either way**, so
nothing currently published depended on the old behaviour — the fix closes a hole rather than
correcting live data.

Gate: 0 CS errors / warnings, 7 968 tests green on both TFMs, registry regenerated last and
idempotent on a second pass.

@codex review

Brain2k-0005:
## Round 6: all seven fixed — and two of them were the wrong shape, not the wrong value

**Clamping was the wrong model for the axis.** It stopped the throw and replaced it with a run
of identical final columns: the same date drawn over and over, headers overflowing a track
measured for one column, and the title you spotted reading `Dec 31 – Dec 30, 9999`. An axis
that reaches the end of the calendar is simply **shorter**, so `BuildColumns` truncates now,
and total width, axis end and the title all derive from the columns actually built.

Detecting the saturation needed stepping from the **previous column** rather than recomputing
from the origin — a saturated advance returns `DateTime.MaxValue`, which is a later instant on
the same final day and so compares as genuine progress. That is also why the first version of
the title fix still read backwards.

**Paging** is clamped in both directions. **The fallback view now commits** rather than merely
looking selected — good catch on the null the single-select ToggleGroup emits when you click
the already-pressed item; the hidden view really did come back when resources returned.
**Null is not the empty resource id**, so unassigned events no longer slip through the
pre-expansion filter just to be discarded after.

**The Razor allowance was wrong in both directions**, and both of your reports are the same
missing distinction. A `.razor` file is scanned as what it is now: **markup regions alternating
with C# code regions** (`@code`, `@functions`, `@{ }`). In markup, a quote opens an attribute
value only **after an `=`** — which is also what keeps an apostrophe in ordinary text
(`<p>don't</p>`) from swallowing the line — and the value's text is not code except for the
`@`-expressions in it. In a code region those same quotes are ordinary C# literals whose
content is text throughout, so `@code { const string E = "@(typeof(Lumeo.Sheet))"; }` no longer
counts. Five cases pin the matrix.

Regenerating moves **no association either way**, so nothing published depended on any of this.

Gate: 0 CS errors / warnings, 7 973 tests green on both TFMs. The earlier `build-and-test` red
was `GanttV3KeyboardNavigationTests.Mid_Flight_Combination_Drift` — the known flake in #413,
unrelated to this PR.

@codex review

Brain2k-0005:
## Round 7: all eight fixed — one of them a real bug that predates this PR

**The recurrence budget counted from DTSTART, not from the window.** CodeRabbit is right, and
the consequence is worse than truncation: a daily standup defined in **1970 needs 20 454
candidates to reach 2026** and rendered as *nothing at all* — in every view, not just this one.
Raising the ceiling in round 2 only moved where the cliff fell.

`ExpandRule` seeks past the run-up arithmetically now and applies the budget from there. The
seek is a deliberate **underestimate** (a whole period is subtracted): starting early costs
iterations, starting late drops a real occurrence — and a booking that began before the window
but runs into it is exactly what a resource view must not lose. Rules with `COUNT` are excluded,
because COUNT is measured over the raw generated series and its position has to be *counted*,
not computed; such a series is small by construction. Three tests, and the disable-check
confirms both directions.

**Markup text is text.** Both of your matcher findings are the same missing distinction, and it
was a live defect: a reference in markup is a **tag** — `<Sheet />`, `</Sheet>` — never prose,
and an inline `@<...>` template inside a code block steps back into markup for the length of the
fragment. Your Button example does not actually associate, but the mechanism does:
**`toggle.json` currently claims `CollapsibleTests.razor`** purely because "Toggle" is the
caption on a trigger. Regenerating drops exactly that one association and nothing else.

**My two char-literal tests were vacuous** — thank you, that is a fair hit. Both asserted
`False` against fixtures that never contained the component name, so no scanner behaviour could
fail them. Rewritten to assert what they were meant to: that the literal is consumed *exactly*,
leaving the reference after it visible.

Also fixed: bars honour `SchedulerEvent.Url`; the minimum bar width no longer pushes a bar past
the track; the saturated title names the last visible **day**; and the track is as wide as the
columns it draws. On that last one I did not take the suggested route — dropping the incomplete
column would delete the caller's own anchor and leave an empty axis, so the track is measured
in whole columns instead, which is what the headers and rules paint.

Gate: 0 CS errors / warnings, 7 987 tests green on both TFMs, registry regenerated last and
idempotent.

@codex review

Brain2k-0005:
## Round 8: all five fixed — and the matcher finding was three live defects, not one

**Attribute names.** You are right that keeping the whole tag kept parameters too, and the
published registry had **three** associations that existed for no other reason:

| entry | claimed | because of |
|---|---|---|
| `label` | `MegaMenuDisabledHost.razor` | `<MegaMenuItem Label="Products">` |
| `text` | `SplitButtonRotationTests.razor` | `Text="Save"` |
| `collapsible` | `SplitterCollapseClampTests.razor` | `<SplitterPane Collapsible="true">` |

The element name is still kept — that is where a reference in markup lives — as is an
`@`-expression used as an attribute value. Regenerating drops exactly those three and nothing
else.

**Control directives** now carry their condition as code; the body stays markup.

## The seek from round 7 had three holes, all yours

- **Weekly seek counted raw `ByDay` entries** while `WeeklyCandidates` collapses them with
  `Distinct()`. A list naming Monday twice — reachable through the legacy `DaysOfWeek` pair —
  sought about twice as far as it should.
- **A reversed event took down the render.** `SchedulerEvent` does not validate `End >= Start`,
  and such data used to be laid out or discarded harmlessly; subtracting a negative duration
  during the seek reached *forward* and threw. That one is squarely on me — the seek turned
  malformed input from a cosmetic problem into a page-level crash.
- **The budget ignored the lookback the seek deliberately keeps.** Measured from the actual seek
  point now.

Gate: 0 CS errors / warnings, 7 990 tests green on both TFMs, registry regenerated last and
idempotent. Previous round's CI was fully green including e2e.

@codex review

Brain2k-0005:
Round 9: all four fixed.

**The today marker** is the one that would have shown up in normal use — thank you. The wrapper resolved the browser date for its own Today button but handed the view `null`, so on a Blazor Server circuit spanning midnight the axis moved and the marker did not. Resolved once interop is available; `null` until then stays the documented prerender fallback, and the test disable-checks in both directions.

**The three matcher findings are all false NEGATIVES** — a parenthesis inside a literal truncating a condition, a `<br>` swallowing an inline template, and a file directive keeping only its keyword while the type was blanked. Both expression forms share one literal-and-comment-skipping balancer now. Regenerating moves **no association at all**, which is the expected signature: these close holes rather than correcting published data, unlike the three false positives in round 8.

Gate: 0 CS errors / warnings, 7 991 tests green on both TFMs, registry regenerated last and idempotent.

@codex review

Brain2k-0005:
## Round 10: all four fixed — the first one made the previous round's fix inert

**The browser date came from the Gantt's module.** `Lumeo.Scheduler` references only the core
package, so in the documented standalone install `./_content/Lumeo.Gantt/js/gantt-v3.js` 404s,
`ResolveBrowserTodayAsync` swallows it, and the marker falls back to the **server's** date —
precisely the scenario round 9 was meant to fix. `scheduler-views.js` answers it now through a
Scheduler-owned interop method, and a test pins the wrapper to it so the Gantt call cannot creep
back.

**Today refreshes the cache** rather than reading the first-render value, so a circuit that stays
connected across local midnight no longer moves the axis while handing the view yesterday.

**Two more matcher false negatives**: an implicit expression's invocation arguments (the
scanner's own contract already promised `@Foo.Bar(...)`), and a `>` inside a quoted attribute
being taken as the tag's end.

## Two things my gate should have caught before CI did

- The previous push went red on **RS0050** — `PublicAPI.Unshipped.txt` still recorded
  `OnAfterRenderAsync` as `*REMOVED*` while I had just re-added it. My build gate filtered on
  `error CS`, and analyzer IDs are not CS. Fixed, and the gate now counts every
  `: (error|warning) [A-Za-z]+[0-9]+:` plus the summary line.
- This round's first build raised **RS0016** and **CS1734**: the new interop member needed
  public-API entries, and I had inserted it between an existing doc comment and the method it
  documents, orphaning that comment onto it. Both fixed before pushing.

Gate: `0 Error(s) / 0 Warning(s)` on a clean rebuild, 7 992 tests green on both TFMs, registry
regenerated last and idempotent.

@codex review

Brain2k-0005:
Round 11: both fixed, and the first test needed a second attempt.

**Tag ends** now step over embedded Razor expressions, so a `>` inside an expression's own string is no longer read as the tag terminator. **Implicit expressions** consume their whole suffix chain — member, invocation and indexer — instead of stopping after the first `()`.

Worth flagging, because it is the second vacuous test I have caught this way: my first version used `Render<Lumeo.Sheet>()` as the swallowed reference and **passed with the fix disabled**. The reason is my own markup model — the swallowed text still contains `<Lumeo.Sheet>`, which reads as an element NAME and is kept. Rewritten around `typeof(Lumeo.Sheet)`, which has no angle brackets, it goes red without the fix as intended. Both suffix-chain tests disable-check cleanly too.

Gate: `0 Error(s) / 0 Warning(s)` on a clean rebuild, 7 992 tests green on both TFMs, registry idempotent and unchanged (both fixes are false negatives, so no association moves).

The previous run's `e2e` red was `GanttV3WheelZoomTests.CtrlWheel_Anchors_The_Zoom_On_The_Pointer` again — second sighting of the same flake, unrelated to this PR, passing on re-run. Recorded on #390.

@codex review

Brain2k-0005:
## Round 12: all eight fixed — two of them published false coverage

**False positives**, both keeping display text as code: an HTML comment's contents
(`<!-- use <Sheet /> here -->` read the sample as a live element) and the suffix chain reaching
across whitespace (`@Get() (Sheet)`), which Razor treats as the end of the implicit expression.

**False negatives**: conditional access in a member chain, comments inside an indexer, a quoted
attribute spanning lines, `<` as a comparison inside a body-level expression, and a generic
component's type parameter — `<Grid TItem="Lumeo.Sheet" />` really is a type reference, matched
by the `@typeparam` naming convention since nothing here knows a component's declared parameters.

Regenerating moves no association, so nothing currently published depended on any of them.

## One fix ships without a test, deliberately

The interpolation-hole traversal in `EndOfString` has **no** regression test. I tried three
fixtures — inside an attribute expression, inside plain C#, and as a false-positive
discriminator — and **all three passed with the fix reverted**, because that traversal only
affects skip paths whose difference the downstream blanking absorbs. The other eight tests do
fail when reverted.

I would rather say that than ship a test that cannot fail; this is the third vacuous test I
have caught in this PR by disable-checking, and the second was mine in the previous round. If
you have a concrete input where the outcome differs, I will pin it.

Gate: `0 Error(s) / 0 Warning(s)` on a clean rebuild, 7 992 tests green on both TFMs.

Previous run: `e2e` green; `preview` failed on a GitHub 429 while downloading
`actions/setup-dotnet` — infrastructure, re-run.

@codex review

Brain2k-0005:
Re-requesting: the trigger at 14:50Z drew no response for ~35 minutes, and GitHub was returning
429/503 in that window, so it may not have registered.

Head `93e0ec7`, CI fully green — build-and-test, e2e, build-docs, preview, mcp-server, bom-check.

Also verified on the deployed preview rather than only in bUnit: zero page errors, the resource
timeline renders 10 columns and 6 bars, `dir="ltr"` sits on both the track **and** its scroll
container, the track measures 960px for 10 × 96px columns, and no bar exceeds it (max right edge
864px).

@codex review

Brain2k-0005:
## Round 13: all four fixed — and the first one is the bill for last round's untested change

The interpolation-hole traversal I shipped **without a regression test** applied to *every*
string, not only interpolated ones. A brace in a plain literal — `Check("{")` — opened a phantom
hole, the closing quote then read as a nested literal, and the scan ran to the end of the input,
blanking the expression around it.

I said last round that I could not construct a case where that change mattered. It mattered in
the opposite direction, and your example is exactly the case I failed to find. It is pinned now,
and it fails with the fix reverted.

**Two more false positives**, both keeping data as code:

- A raw-text element's body is character data — Razor does not instantiate markup inside
  `<script>`, `<style>`, `<textarea>` or `<title>`.
- A Razor comment inside an inline template: `@*` made no progress through the expression
  scanner, so a sample tag in the comment counted as a live element and the root never balanced.

**One false negative**: a directive attribute parses its value as C# with no leading `@` on the
value itself, so `<div @key="typeof(Lumeo.Sheet)">` is a real reference.

Four of the six tests fail with the fix reverted; the other two are deliberate over-reach guards
— a component after a `</script>` still counts, and a plain attribute beside a directive one is
still text.

Gate: `0 Error(s) / 0 Warning(s)` on a clean rebuild, 7 992 tests green on both TFMs, registry
regenerated last and unchanged.

Round 12's CI was fully green — build-and-test, e2e, build-docs, preview, mcp-server, bom-check.

@codex review

Brain2k-0005:
## Round 14: all six fixed — every one a consequence of round 13's new state

That is what introducing state does, and four of the six are about raw text alone: a
self-closing `<textarea />` has no body (this repo writes it that way, so the missing end tag
swallowed the rest of the file), `</scripture>` must not close a `<script>`, a Razor transition
inside the body is still evaluated by Razor, and the inline-template scanner needed the same
state or a `<span>` inside a `<script>` kept the template from ever balancing.

Raw-text handling is **one shared helper** now instead of two ad-hoc copies — the duplication is
what let the second site be forgotten in the first place.

## The directive fix was backwards, and its own test caught it

I required a `$`/`@` prefix to tell a nested literal from the attribute's own quote. That is
exactly inverted: a nested C# string is usually **bare**, so
`@onclick="@(() => { Log("x"); ... })"` still terminated at `Log`'s argument. The terminator is
now the delimiter at **bracket depth zero**.

Worth noting how that surfaced: the test failed **at baseline**, not only when reverted — the
disable-check is catching my incomplete fixes, not just my vacuous tests.

Two tests needed rewriting to become discriminating at all:

- one used `Render<Lumeo.Sheet>()`, which survives regardless because the markup model reads
  `<Lumeo.Sheet>` as an element name — the same trap as round 11;
- the char-literal one had to move into a directive attribute, because that is where the literal
  is actually **skipped** rather than blanked by a different path.

All six fail with the fix reverted now.

Gate: `0 Error(s) / 0 Warning(s)` on a clean rebuild, 7 992 tests green on both TFMs, registry
regenerated last and unchanged.

@codex review

Brain2k-0005:
## Round 15: all four fixed — three of them published prose as coverage

`@@` is an **escape**, so `<p>@@typeof(Lumeo.Sheet)</p>` is literal text; blanking one character
at a time let the second `@` open a transition. A brace inside a **comment** is not a hole
delimiter — it held an interpolation hole open past the real closing brace, the string skip ran
past the attribute's own terminator, and the markup behind it was processed as C#. And an end tag
spelled inside a Razor literal is **string data**, so selecting it as a raw-text body's close
resumed markup mode mid-expression.

The fourth is the opposite direction: a Razor control body allows C# **statements** alongside
markup. It is scanned as code with markup islands now — a run starting at `<` goes through the
markup scanner, everything else through the code one — and both halves are pinned: a statement in
the body counts, a caption in it still does not.

One test needed reshaping to discriminate: asserting that prose is *not* published passed either
way, because the runaway blanked the prose too. It now asserts that a real tag **after** the
malformed attribute survives, which the runaway swallows.

Gate: `0 Error(s) / 0 Warning(s)` on a clean rebuild, 7 992 tests green on both TFMs, registry
regenerated last and unchanged.

Round 14's CI: `build-and-test`, `mcp-server`, `bom-check` and `preview` green. `build-docs` and
`e2e` failed twice in **Set up job** on GitHub 429/503 while downloading `actions/setup-dotnet`
— infrastructure, unrelated to the diff; re-run.

@codex review

Brain2k-0005:
## Round 16: all eleven fixed — and the duplication that caused them went first

Half of these said *"fresh evidence beyond the fix in round N"*, and that diagnosis was the
useful part. The same construct was being scanned in several places, each with its own partial
handling, so a fix landed in one and the others kept the bug:

- escapes were handled in markup text but not in attribute values;
- raw-text close discovery knew about expressions but not comments or code blocks;
- the control-body island balancer repeated the inline-template scan without its transition
  handling.

So the duplication went first, before any of the individual fixes:

- **`EndOfTransition`** is now the one place that knows what a Razor transition is — `@@`,
  `@* *@`, `@{ }`, `@:`, `@(...)`, an implicit chain. Every scanner that walks markup calls it.
- **`EndOfElementSubtree`** replaces the two near-identical element walkers (inline templates and
  control-body islands) — which is precisely why each of their fixes had to be made twice, and
  why the second site kept getting missed.

On that base: the null-forgiving `!` and implicit `await` continue an expression; a control body's
braces count as structure only outside literals and comments; `else` / `else if` / `catch` /
`finally` continue the construct instead of handing the branch back to the markup scanner; `@:`
is a markup island; an attribute value may wrap onto the next line; and a component parameter can
carry an expression with no leading `@` — this repo writes `NotFoundPage="typeof(Pages.NotFound)"`
— so a value that **cannot** be display text is scanned as code. That last rule is deliberately
narrow: `typeof(...)` only, with a guard test pinning that `ComponentType="Lumeo.Sheet"` stays
text.

11 of the 12 tests fail with the fix reverted; the twelfth is that narrowness guard.

Gate: `0 Error(s) / 0 Warning(s)` on a clean rebuild, 7 992 tests green on both TFMs, registry
regenerated last and unchanged.

Round 15's `e2e` red was `GanttV3WheelZoomTests.CtrlWheel_Anchors_The_Zoom_On_The_Pointer` for the
third time — unrelated to this PR, recorded on #390, green on re-run.

@codex review

Brain2k-0005:
## Round 17: one finding, fixed — and it is the round-10 failure mode arriving by another route

`SchedulerViewsGetLocalDateAsync` calls an export `scheduler-views.js` only gained in this
release, and the module was imported without the `?v=` query the core modules use. A browser
holding the previous copy finds `getLocalDateIso` missing, the call rejects, and the marker falls
back to the **server's** date without saying so — exactly what the Gantt-module dependency did two
rounds ago, just reached differently. `AppendVersion` already exists for this and already covers
`_content/Lumeo*`; the import simply bypassed it.

Its test needed a second attempt: anchoring on the first mention of the file name landed on an
earlier **comment** that names it, so the assertion inspected a line that could never carry the
call. It matches the import expression exactly now and also asserts the unversioned form is
absent — verified in both directions.

Not changed, and worth saying rather than silently widening: `motion.js`, `gantt-v2.js` and
`gantt-v3.js` bypass `AppendVersion` the same way. That is pre-existing and none of them gained a
new export here, so I left them alone rather than changing cache behaviour for three other
components inside this PR.

Gate: `0 Error(s) / 0 Warning(s)` on a clean rebuild, 7 993 tests green on both TFMs, registry
regenerated last and unchanged.

Round 16's CI was fully green — build-and-test, e2e, build-docs, preview, mcp-server, bom-check,
CodeRabbit.

@codex review

Brain2k-0005:
Re-requesting for head `33c2b99`. The request at 16:58Z drew no response in ~25 minutes, and
GitHub has been returning 429/503 across the API and Actions in this window — three CI jobs failed
in *Set up job* on it earlier, and several of my own API calls were rejected — so it may not have
registered.

State on `33c2b99`:

- All six checks green: build-and-test, e2e, build-docs, preview, mcp-server, bom-check.
- 7 993 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- Registry regenerated last and unchanged.

The previous round's review produced no inline findings on `46468e8`, so if that was a pass, the
only change since is the one-line cache-busting fix and its test.

@codex review

PR #425, opened 2026-08-17
chore(scheduler): remove what the FullCalendar wrapper left behind

Brain2k-0005:
The wrapper is gone and the docs page issues no third-party request, but the library
was still **declared, described and shipped** in four places that outlive the code.

## What was still there

| where | what it said | who reads it |
|---|---|---|
| `Lumeo.Scheduler.csproj` | "powered by FullCalendar", `fullcalendar` tag | **nuget.org** — published, outlives any repo-only fix |
| `Lumeo.csproj` | "Scheduler (FullCalendar)" | nuget.org |
| `RegistryGen/Program.cs` | "Calendar/agenda scheduler wrapping FullCalendar." | docs catalog, search index, **MCP server**, skill catalog |
| `docs/wwwroot/index.html` | importmap for five packages + vendored `preact` | every docs page load |

The vendored FullCalendar directories went with the wrapper, so the importmap pointed at
paths that 404. `preact` was vendored **only** as FullCalendar core's own renderer, and
nothing else references it — 3 files, 23 KB, dead.

## A limitation that had already been fixed

`SchedulerRRuleParser` still documented that `<Scheduler>` ignores a `Recurrence` rule,
because its serializer branched on the legacy `DaysOfWeek` pair alone. It renders the
first-party views now, so the rule is honoured everywhere. A reader following that
paragraph would have avoided a feature that works.

`SchedulerView` also documented itself as mapping onto `dayGridMonth` / `timeGridWeek` /
`timeGridDay` / `listWeek`, and `SchedulerEvent` called itself the contract between Blazor
and the JS wrapper.

## What is deliberately kept

Comments that record **why an API is shaped the way it is** — exclusive end timestamps,
"widest bar first" lane ordering, the toolbar's extraction from the wrapper. Those are
statements about the past and stay true.

That is why `SchedulerDependencyGuardTests` is not a blanket token ban like
`IconDependencyGuardTests`. It checks the four places a *real* dependency reappears — a
loaded asset under the package's `wwwroot`, a declared CDN dependency, a claim in published
package metadata, a vendored copy on the docs site — and matches reference forms (an npm
specifier, a vendored path, the camelCase CDN key) rather than the bare word. A returning
historical note stays legal; a returning import does not.

Disable-check: reverting any one of the four changes fails exactly its own test.

## Not in this PR

The core package description still claims "4,831 unit tests" (7,917 today) and "164
components" (165). Left alone deliberately — a count in published metadata drifts on every
PR, and deciding whether it should be maintained at all is a separate call.

Gate: clean rebuild 0 errors / 0 warnings, 7,917 tests green on both TFMs, registry and
skill catalog regenerated last.

@codex review

## Summary by CodeRabbit

- **New Features**
  - Scheduler documentation now highlights calendar, agenda, resource, and multiple time-range views.
  - Recurrence rules are documented as supported across all scheduler views.
  - Scheduler descriptions now reflect its first-party calendar engine.

- **Bug Fixes**
  - Removed obsolete calendar integration assets and configuration.

- **Tests**
  - Added coverage ensuring the Scheduler operates without the removed calendar dependency.


Brain2k-0005:
Both fixed, and the second one found a defect in my own fix.

**The guard now covers RegistryGen's summary map.** You are right that it was the gap that
mattered: package metadata under `src/` was checked, but the one string feeding the docs
catalog, the search index, the MCP registry and the skill catalog was not — restoring
"wrapping FullCalendar" there passed every test in the class.

Writing that check, its **disable-check failed to fail**: anchoring on the first
`["Scheduler"] = "..."` entry in `Program.cs` finds the *package assignment*
(`= "Lumeo.Scheduler"`), not the summary, so the guard was inspecting a line that can never
carry the claim. It now checks every such entry. Verified both directions: green as
committed, red with the old text restored.

**Timeline claim dropped** from the summary and the package tags. The view lands in #424;
advertising it here points users at something they cannot use.

Gate: 0 CS errors / warnings, 7 918 tests green on both TFMs, registry, search index and
skill catalog regenerated last and idempotent on a second pass.

@codex review

Brain2k-0005:
All three fixed.

**Paths as well as contents** — you are right that a minified UMD bundle need not contain any of the reference forms, so the relative path is now checked alongside the file. **preact** is matched as a vendor *path* (`lumeo-vendor/preact`) in both the host page and a full recursive scan of the vendor root, rather than as the bare word, so a future first-party use of preact would not trip the guard.

Verified by planting each shape — a nested `preact.mjs`, and a `fullcalendar-shim.js` whose contents are innocuous — and confirming the matching test goes red, then green once removed.

**Recurrence-support list** corrected: `SchedulerResourceView` expands `Recurrence` too.

Gate: 0 CS errors / warnings, 7 918 tests green on both TFMs, registry regenerated last and idempotent.

Brain2k-0005:
The 👍 above predates `f714727`, which widened the guard to file paths and vendored `preact`, and corrected the recurrence-support list. Requesting a look at that commit specifically.

@codex review

PR #426, opened 2026-08-18
feat(scheduler): add calendars, panes and an opt-in appointment dialog

Brain2k-0005:
Closes the two remaining 5.0.0 items: calendars with overlay/side-by-side panes, and the opt-in
appointment dialog.

## Calendars are not resources

A `SchedulerResource` is what an event **consumes**; a `SchedulerCalendar` is which feed it
**came from**. They stay separate types because they answer separate questions — a resource
decides *where* an event sits, a calendar decides *whether it is shown at all* — and an event can
carry both: Alice's booking, from the Team calendar.

Supplying `Calendars` renders visibility chips. The chip keeps its colour dot when switched off,
at reduced opacity, because that dot is the only thing tying it to its events.

**Null and the empty string are different**, the same distinction `ResourceId` already draws. A
null `CalendarId` means the event belongs to no calendar, so no toggle owns it and nothing can
hide it — otherwise switching off "Team" would silently swallow everything unfiled, with no way
to get it back. A blank id is a real calendar a caller can declare.

## Panes

`PaneMode` chooses whether visible calendars share one view or each get their own. Side by side
answers the question overlaying cannot — *what does each look like on its own* — so it needs at
least two visible calendars to mean anything. With one it is the overlay wearing a column header;
hiding the last one falls back to overlay rather than rendering an empty box.

The pane header's control merges **all** panes rather than just its own: the mode is not a
per-calendar flag, and a control that looked per-pane would promise something it cannot deliver.

Every pane renders the **same view fragment**, not a copy of the switch, so a pane cannot drift
from the overlay the first time a view gains a parameter. A selection drawn in a pane reports that
pane on `SchedulerDateRange.CalendarId`, so a caller with its own form can file the new event into
the column the user drew it in; `null` means the calendars are overlaid and no single one owns the
gesture.

## The dialog never displaces your own workflow

That is the whole design constraint, and it is what most of its tests pin. `BuiltInEventDialog`
is **off by default** and, when off, renders nothing at all. Turning it on leaves `OnDateSelect`
and `OnEventClick` firing exactly as before — the dialog is *additive*, so a caller keeps whatever
it already hangs off them. Saving reports through `OnEventChange` + `EventsChanged`, the pair a
drag already uses, so nothing downstream learns a new channel.

A new event lands in the first **visible** calendar, not the first declared one: creating into a
switched-off calendar would make it vanish the moment it was saved.

Save and delete resolve the edited event through the **same unambiguous resolver the drag path
uses** — extracted, not re-implemented. A blank or duplicated id identifies nothing, so the stored
collection is left alone and the dialog closes rather than sitting there uncommittable.

## Time zones and all-day

The dialog is populated from the **display** projection, so what it hands back is read through the
same projection a drag uses — otherwise a title-only edit rewrote the caller's data into a
different frame. All-day events are the exception at both ends: `ToDisplay` never projected them,
so the form showed them raw and they are committed raw, as naive bounds. And an all-day end is
**exclusive**, so a single day is stored as ending at the next midnight — collapsing both bounds to
one midnight produced a zero-length event that disappeared the moment it was saved.

## Verification

- 8 032 tests green on both TFMs (39 new), `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 33 Scheduler E2E tests green, 4 of them new: the wrapper's own chrome had no real-browser cover
  at all before this, only the view components one level below it. `/e2e/scheduler-shell-preview`
  drives the public component through the create/edit/delete round trip, panes splitting and
  collapsing back to the overlay, and the dialog staying out of the way when it is off.
- Every new test disable-checked one fix at a time; the guards that stay green with the feature
  reverted are deliberate (no chips without calendars, resource colour still beats calendar, the
  caller's handler fires with the dialog off, an overlaid selection carries no calendar).
- 10 new strings across all 13 locales.

## Summary by CodeRabbit

* **New Features**
  * Added an optional built-in appointment dialog for creating, editing, validating, and deleting events.
  * Added named calendars with visibility controls, color support, and overlay or side-by-side layouts.
  * Added calendar-aware event selection, filtering, and pane-specific interactions.
  * Added Scheduler demonstrations and an end-to-end preview page.

* **Bug Fixes**
  * Prevented month-view popover ID conflicts when multiple schedulers are displayed.

* **Localization**
  * Added Scheduler calendar and event-management translations across supported languages.


Brain2k-0005:
CI caught something my local gate structurally cannot, so recording the diagnosis before the next run.

**A real regression, fixed in `be99abb`.** The views arm their drag-create gesture on `OnDateSelect.HasDelegate` — deliberately, so a gesture that cannot act is never offered. Routing every view through the dialog's internal handler armed it unconditionally, including for callers with no handler and the dialog off. Now a delegate is supplied only when the caller wants the range or `BuiltInEventDialog` is on, pinned by three tests.

**The failing E2E test is a separate question, still open.** `SchedulerViewsDragTests.Month_Drag_Move_Creates_A_Ghost_Then_Removes_It_And_Commits` targets `/e2e/scheduler-views-preview`, which uses `SchedulerMonthView` directly — not the wrapper this PR changes. I reproduced the failure locally, then ran the same test against **master** in the same worktree: it fails there too. So my local reproduction proves nothing about this branch; it is the known Windows/Chromium drag discrepancy.

In CI the picture is the opposite: master passed it on the last three runs, this branch failed it once. One data point each, on a test driven by a synthetic mouse drag. This push re-runs it.

`build-and-test` failed on `GanttV3KeyboardNavigationTests.Mid_Flight_Combination_Drift` — the known flake in #413.

Gate: `0 Error(s) / 0 Warning(s)`, 8 022 tests green on both TFMs.

Brain2k-0005:
Both reds explained, and the interesting one was not this branch.

## The e2e failure is a date-dependent test bug

`Month_Drag_Move_...` hard-codes its drop target to **day 20**, while the preview seeds `e2e-1` at `DateTime.Today.AddDays(2)`. On the **18th of any month** those are the same day, so `CommitDrag` hits its `newStart == ev.Start` guard and returns without committing — while every ghost assertion before it passes, which is exactly what made it read as a drag-engine regression.

What settled it: master passed this test the day before, so the branch looked guilty. Running the same test **against master in this worktree** reproduced the failure — that ruled the branch out and left the date as the only variable. Fixed at the cause: the target is derived from the event's own cell now, falling back to day 21. Verified green on the date that broke it — all 8 drag tests pass locally.

It would have failed CI on the 18th of every month, on any branch.

## The other red was mine

`build-and-test` went red on **registry staleness**: the gesture-fix commit added three tests, which changes the per-component test count, and I did not regenerate afterwards. Regenerated — that is the "registry last" rule, and I broke it by committing a test fix after the last regeneration.

## Also in this push

The gesture-arming fix from `be99abb` — a genuine regression this PR introduced, where routing every view through the dialog handler armed the drag-create gesture for callers that had no handler and no dialog.

Gate: `0 Error(s) / 0 Warning(s)`, 8 022 tests green on both TFMs, registry regenerated last and idempotent, and the Scheduler drag e2e suite run locally against this branch's own docs build.

@codex review

Brain2k-0005:
## Round 1 — all four applied, each with a test that discriminates

**1. A title-only edit shifted the time.** The dialog is populated from the display
projection, so its wall-clock values have to be read back through the same projection the
drag path already uses. Saving stored the shifted reading as an `Unspecified` instant, which
rewrote the caller's data into a different frame on every save. Now it round-trips through
`SchedulerTimeZoneProjection.FromDisplay`. The delete path had the same shape of bug for a
different reason — it matched the victim by reference, and the dialog holds the projection,
a different instance — so it matches by id now.

**2. Creating in a pane filed it into the wrong pane.** Every pane shares one handler, so
the pane's own calendar id now travels with the callback rather than being re-derived as
"first visible calendar".

**3. All-day left the instants it hides.** The controls become date-only, so values behind
them were saved unseen and reappeared on toggling back. The toggle normalises to `.Date`.

**4. Overflow popover ids collided across panes.** The id also keys the global click-outside
registry, so two month views on one page meant opening a popover in the second re-registered
the first one's id against the second callback: the wrong one closed, and the other could not
be dismissed at all. Per-instance ids now.

## Verification

- 8 026 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- Disable-check per fix, one at a time: each of the four new tests goes red with only its own
  fix reverted. Two of them were wrong on the first write and caught here — one drove a
  `Switch` as if it raised `onchange`, the other looked for popovers in a month with no
  overflowing day, so it asserted `NotEmpty` against a collection that could never be
  populated.

@codex review

Brain2k-0005:
## Round 2 — all four applied

**All-day → timed shifted the date.** `ToDisplay` deliberately leaves all-day instants alone, so
they were never projected and inverting one moved the event a day. The save now skips the
round-trip when the source event was all-day.

**All-day ends are exclusive**, which the month layout and the expander's `End > rangeStart` check
both read — so collapsing both bounds to the same midnight produced a zero-length event that
disappeared the moment it was saved. One helper now normalises to at least one full day, and it
covers the other route to the same value: a user picking the same date twice in the form.

**Delete on a duplicated id removed every record sharing it.** That is the same ambiguity the drag
path already refuses to act on, so rather than re-implement the rule I extracted the drag path's
resolver and pointed all three call sites at it — save had the same first-match bug, just without
the data loss.

**`SchedulerDateRange` carries its pane.** With the built-in dialog off, a caller running its own
creation workflow saw one selection shape from every pane. It is a body property, not positional,
so the primary constructor and `Deconstruct` are untouched; `null` still means "overlaid, no single
calendar owns this gesture", and a test pins that too.

## Also in this round

The wrapper's own chrome had **no real-browser coverage at all** — the existing harness drives the
view components directly, one level below the calendars, panes and dialog. A second harness at
`/e2e/scheduler-shell-preview` drives the public component, with Playwright tests for the
create/edit/delete round trip, panes splitting and collapsing back to the overlay, and the dialog
staying out of the way when it is off. That last one is the design constraint of the whole feature,
so it is worth having a browser assert it and not just bUnit.

## Verification

- 8 031 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 33 Scheduler E2E tests green against a fresh Release docs build.
- Disable-check per fix, one at a time: each of the five new bUnit tests goes red with only its own
  fix reverted, and the four new E2E tests all go red with the harness opted out.

@codex review

Brain2k-0005:
## Round 3 — the flip side of my own round-2 fix

You are right, and it is the same event caught from the other end. Skipping the inverse
projection preserved the **date**, then left a UTC kind on an event that is no longer all-day —
so the next render projected it and drew 00:00Z at 01:00 Berlin.

The form showed those bounds raw, because `ToDisplay` leaves all-day instants alone. Naive bounds
are what those readings were all along, so that is what is stored now. The test asserts both
halves: the kind is `Unspecified`, and projecting the committed value into the display zone
returns it unchanged — which is the property that keeps it from moving on the next render.

## Also in this round, from re-reading my own change

Saving against a duplicated id left the collection untouched but returned with the dialog **open**,
so the user was pressing Save against a form that could never commit, with nothing to tell that
apart from a slow render. It closes now, and a test pins it.

And the calendars section of the docs now says that a selection drawn in a pane reports that pane
on `SchedulerDateRange.CalendarId`, since that is the whole point of exposing it.

## Verification

- 8 032 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 33 Scheduler E2E tests green against a fresh Release docs build.
- Disable-check: both new tests go red with only their own fix reverted. The duplicate-id probe
  needed two reverts to be honest — reverting the resolver alone still left the guard standing, so
  the test passed and looked non-discriminating until the guard came out with it.

@codex review

Brain2k-0005:
## Round 4 — both applied

**Every unresolved edit is now non-committable, not only the duplicated one.** The reachable case
is a parent dropping the event while the dialog stands open on it — a poll landing, a sibling
deleting it. The old guard let that through to report the record on `OnEventChange` while
`EventsChanged` carried the unchanged collection: two notifications contradicting each other about
an edit that had gone nowhere. Creations are untouched.

Worth noting for the blank-id half of that finding: a blank id never reaches the dialog, because
the views do not draw an event with one. The test therefore pins the case that *is* reachable
rather than one that reads well and cannot happen.

**The split control follows what is visible.** With everything but one calendar switched off there
is nothing to put beside anything, `ShowsPanes` refuses it, and the control was promising a state
it could not reach. The test also checks the other direction — switching the second calendar back
on brings both the panes and the control with it.

## Also in this round, from re-reading my own round-2 change

The one-day floor for all-day ends applied to **every** all-day save, so it re-dated an end a
caller had already set: an all-day event carrying a same-day afternoon end, which the layout reads
as one day either way, came back re-dated from an edit that never touched it. That is the same
defect as shifting a start, just at the other bound. The floor now applies only when the end is not
after the start — which is the case it existed for.

## Verification

- 8 035 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 33 Scheduler E2E tests green against a fresh Release docs build.
- Disable-check per fix, one at a time; all three go red with only their own fix reverted.

@codex review

Brain2k-0005:
## Round 5 — all three applied

**A cleared date field no longer commits what it hid.** The parser keeps the previous value on
unparseable input, and that is deliberate — it is what lets a field be retyped without destroying
itself mid-keystroke. But it made clearing the field indistinguishable from not touching it, so
Save committed a date the form no longer showed. Save now knows about more than a blank title, and
the test checks both directions: disabled while the field is empty, released once a date is typed
back in.

**`Selectable="false"` keeps the create dialog shut.** The views gate that gesture on
`OnDateSelect.HasDelegate` alone, so arming it for the dialog armed it past the parameter that had
turned selection off. A caller's own handler is left exactly as it was — that contract predates the
dialog and changing it here would be a separate decision.

**Panes are keyed by calendar, not by position.** Good catch on the reorder case — the state that
follows the wrong feed is the overflow popover, which then lists another calendar's events while
looking like it belongs where it was opened. The test opens the popover in the first pane, has the
parent hand the same two calendars back the other way round, and checks the popover is still the
same pane's.

## Verification

- 8 038 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 33 Scheduler E2E tests green against a fresh Release docs build.
- Disable-check per fix, one at a time; all three go red with only their own fix reverted.

@codex review

Brain2k-0005:
## Round 6 — all five applied

The P1 is the one worth naming: **the dialog is a snapshot of a world that keeps moving.** Saving
rebuilt the event from the copy the form opened with, while the parent's refresh had already been
adopted into the working copy — so a title-only save reverted colour, recurrence and extended
props, none of which the form even shows. It builds on the current stored record now, and the
projection decision still reads the snapshot, because *whether these numbers were projected* is a
fact about the moment the form was filled, not about the record today.

The zone finding is the same shape and got the same treatment: the dialog captures the zone its
numbers were read in, so a `TimeZone` change underneath an open form no longer reinterprets digits
nobody retyped.

The remaining three:

- Toggling all-day re-renders both date controls from the retained bounds, so the validity flags
  reset with them rather than keeping Save disabled over a form with nothing wrong.
- A calendar removed underneath the dialog keeps an option of its own, so the select shows what the
  field actually holds instead of falling back to "no calendar" over a live value.
- A repeated calendar id draws one pane instead of throwing. Keying the panes by id is what made a
  duplicate fatal, so the dedupe belongs with it — first entry wins, and the caller's order
  survives.

## Verification

- 8 043 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 33 Scheduler E2E tests green against a fresh Release docs build.
- Disable-check per fix, one at a time; all five go red with only their own fix reverted.

@codex review

Brain2k-0005:
## Round 7 — applied

Only the dialog's **markup** was conditional, so the state outlived the opt-in and came back with
it. Cleared now when the feature is withdrawn — and the create form goes the same way when
`Selectable` is withdrawn, since that is the gesture that opened it. An edit survives that one:
clicking an event has nothing to do with selection, and a second test pins the distinction so a
later tidy-up cannot quietly collapse the two.

Worth noting what the assertion had to become. A freshly mounted `Dialog` with `Open="false"`
renders no content at all, so the "it did not reappear" check is the absence of the form, not an
`aria-modal="false"` on it — my first version asserted the latter and failed against the fixed
code, which is the more useful way round for a test to be wrong.

## Verification

- 8 045 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
  One run of `GanttV3KeyboardNavigationTests.Mid_Flight_Combination_Drift` failed on net8.0 and
  passed on net10.0 in the same pass, then passed in isolation on both — that is issue #413, a
  known load-sensitive flake in code this branch does not touch.
- 33 Scheduler E2E tests green against a fresh Release docs build.
- Disable-check: both new tests go red with the reset removed.

@codex review

PR #427, opened 2026-08-18
fix(scheduler): give the time grid its day header, and English its dialog strings

Brain2k-0005:
Two defects reported against the live docs site.

## The week view looked cut off because its header was never there

Week and Day render through `SchedulerTimeGridView`, which went straight from the live region to
the all-day strip to the grid. Nothing on screen said whether a column was Monday or Thursday —
which is exactly what "cut off at the top" looks like.

Month has carried its weekday row from the start. This is the same row plus the **date**, because
a time grid shows one specific week where the month grid shows a repeating pattern. Both rows read
`AbbreviatedDayNames` from the current UI culture, so the two views cannot disagree about what
Monday is called, and today is marked the way the month grid marks it.

A test asserts the header dates equal the `data-daycol` dates below them — the header is a
separate row carrying its own column definition, which is the one thing that can silently drift.

## The first hour label really was clipped

Every hour label is lifted onto its own gridline by `-top-1.5`. The first has no gridline above
it — only the scroller's clipping edge — so the lift sliced its digits in half. It sits on the top
edge now; the rest keep the lift, because the fix is for the edge case and not a change of
alignment.

## Nine dialog strings were missing from English only

`Scheduler.NewEvent`, `EditEvent`, `EventTitle`, `Starts`, `Ends`, `NoCalendar`, `Save`, `Cancel`
and `Delete` existed in **all thirteen other locales** and not in English. English is the fallback
every unresolved culture lands on, so everyone except German read raw keys off the live site.
`Scheduler.NoEvents` had the same gap the other way round — only en/de had it — and is now
translated everywhere.

### The guard that should have caught it

There is a completeness guard. It works off a hand-maintained array, so it only checks the keys
somebody remembered to add to it, and the keys nobody remembers are precisely the ones that break.
I extended that list, but the real fix is a second guard that **derives** the check: every key any
translation carries must also exist in English. It exempts only the CLDR plural categories English
does not have (`Zero`/`Two`/`Few`/`Many`), and only when the family's `.Other` form *is* in
English, so the exemption cannot swallow a typo in the stem.

Reverting the nine keys makes it fail naming every affected locale and key.

## Verification

- 8 052 tests green on both TFMs (10 new), `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 34 Scheduler E2E tests green, one new: the header lines up with its columns and the first label
  sits inside the scroller — both geometry, which is why that half is Playwright's.
- Disable-check per fix: removing the header row reddens 5 tests, restoring the lift reddens the
  label test, reverting the English keys reddens the derived guard.

@codex review

## Summary by CodeRabbit

- **New Features**
  - Added localized weekday and date headers with current-day highlighting.
  - Improved sticky scrolling, shared pane scrolling, and configurable all-day strip sizing.
  - Added localized Scheduler event-management, empty-state, and overflow-count labels.
  - Enhanced event overflow popovers with keyboard support and RTL-aware positioning.
- **Bug Fixes**
  - Prevented first-hour labels and overflow dialogs from being clipped.
  - Ensured “today” reflects the configured display date and refreshes correctly.
- **Tests**
  - Expanded coverage for headers, scrolling, localization, overflow behavior, and date alignment.


Brain2k-0005:
## Round 1 — all three applied

**The header and the grid now share one scroller.** You are right that the widths could not
agree: the header was laid out against the full component width while the grid lost the scrollbar
to it. Rather than reserve a matching gutter — arithmetic that has to be right on every platform —
the header and the all-day strip moved *inside* the scroller as one sticky block, which makes the
widths equal by construction. It also fixes the same drift in the all-day strip, which sat outside
too, and a header that survives a scroll through 24 hours is what the view wanted anyway.

One sticky wrapper rather than two sticky rows: a second row's `top` would have to equal the first
row's rendered height, and nothing in the markup can measure that.

**And you are right about my test.** Comparing the date arrays passes happily while the two rows
are laid out against different widths — the header names the right day and draws it over the wrong
column, which is precisely the bug. It compares x-coordinates and widths now, with a one-pixel
tolerance for rounding and nothing like a scrollbar's worth.

**Today comes from the resolved zone.** The timeline already takes it from the wrapper for exactly
this reason; the time grid takes the same one now, through a `Today` parameter shaped like the
timeline's — nullable, defaulting to the host's own, so a caller can also pin it for a
deterministic screenshot. The test checks both halves: the projected day is marked *and* the
host's own day is not, which is the half that stays green if the parameter is read but the
fallback left in place.

**The locale files are back to LF.** Twelve one-line additions had become five thousand changed
lines with a carriage return on every rewritten one. The staged diff is now exactly
`12 files changed, 12 insertions(+)` for those files, and `git diff --check` is clean. I also
added a check over the whole branch diff — no staged file's CR count differs from master's.

## Verification

- 8 055 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 34 Scheduler E2E tests green against a fresh Release docs build.
- Disable-check per fix: moving the scroll back onto the inner row reddens the shared-scroller
  test, dropping the sticky classes reddens the sticky test, restoring `DateTime.Today` reddens
  the zone test.

@codex review

Brain2k-0005:
## Round 2 — all four applied

**The all-day strip is capped and scrolls in its own right.** Moving it inside the sticky block is
what made its height dangerous: a lane per overlapping event, and once the block outgrew the
viewport it covered the timed rows for the whole scroll range. That is a defect my own round-1 fix
introduced, so thank you for chasing it there.

**`aria-current="date"` on today's cell.** The marker was colour plus a `data-` attribute, which
announces nothing — a screen reader read today as the same weekday and number as every other
column. The test also checks that exactly one column claims it, since `aria-current` everywhere
announces as little as nowhere.

**The day number follows the culture's calendar.** `DateTime.Day` is always the Gregorian one. The
test asserts the rendered number equals `ToString("%d", ar-SA)` *and* that it differs from
`DateTime.Day` — without that second half it would pass against the very bug it exists for.

**The browser's date is re-read rather than cached.** Skipped entirely when a display zone is
configured, because then nothing reads it; otherwise one small interop per render, which the
scheduler only does on user actions. The test counts the calls and moves the browser's midnight
between two renders.

## One thing your review found indirectly

My first attempt put the `aria-current` note as a Razor comment **inside the opening tag**. The
repo has a lint guard for exactly that — a comment there compiles clean and misbinds attributes at
runtime — and it failed the suite immediately. The comment sits above the element now. Worth
recording because the guard earned its keep.

## Verification

- 8 059 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 34 Scheduler E2E tests green against a fresh Release docs build.
- Disable-check per fix, one at a time; all four go red with only their own fix reverted.
- No line-ending churn: every file in the branch diff has the same CR count as master.

@codex review

Brain2k-0005:
## Round 3 — applied

You are right, and it is the same bug I fixed in round 1, one level down: capping the strip with
its own scroller took the scrollbar's width out of that row alone, so its columns drifted from the
header above and the grid below.

So the scroller is gone. The strip is capped by **lanes**, with the surplus counted per day — which
is what the month grid already does with the very same packer (`SchedulerMonthLayout.PackRow`), and
it reuses the existing `Scheduler.MoreEvents` plural key, so no new strings. Nothing inside the
strip scrolls, so it stays exactly as wide as everything else, and the lane budget matches the
month grid's so a reader moving between the two views does not find one quietly deeper than the
other.

## An older defect the test turned up

Asserting that no lane is drawn past the budget failed with lanes reading `3` — the loop's **exit**
value. The pill's `data-lane` is read from inside `Tooltip`'s `RenderFragment`s, which are
evaluated after the `for` loop has finished, so every pill carried the exit value rather than its
own lane. Uncapped, that meant `data-lane="24"` on a 24-lane day.

The index is copied per iteration now, the same way `SchedulerMonthView` copies its day index —
and the comment says why, because the next person to write `@lane` inside a fragment there will
have no reason to suspect it.

## Verification

- 8 059 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 34 Scheduler E2E tests green against a fresh Release docs build.
- Disable-check: removing the lane cap and restoring the loop-variable read each redden the test
  on their own.
- No line-ending churn: every file in the branch diff has master's CR count.

@codex review

Brain2k-0005:
## Round 4 — both applied

**The overflow is reachable.** Correct: capping the lanes takes the surplus out of the DOM, so a
count with nothing behind it made the fourth and later appointments visible as a number and
openable by nothing. It is a button with a popover now — the same self-contained popover
`SchedulerMonthView` uses for exactly this overflow, including the per-instance id that keeps two
grids on one page from dismissing each other's registration, which is a lesson from PR #426 rather
than something I would have got right by symmetry.

Two tests: the popover lists precisely the events the lanes could not hold and clicking one reaches
`OnEventClick` and closes it; and two grids rendered side by side get distinct popover ids.

**`Scheduler.MoreEvents` is translated everywhere** (thanks CodeRabbit). It existed only in en/de,
so every other locale fell back to English for a label the **month grid** has been drawing all
along — my reuse in the time grid widened an existing gap rather than creating one. All twelve
locales have it now, with the `.Few`/`.Many` forms `PluralKey` selects for Russian, Polish and
Arabic.

The base, `.One` and `.Other` forms went into the hand-maintained completeness guard, so a new
locale has to translate them. Deliberately *not* `.Few`/`.Many`: those are Slavic and Arabic
categories that English and the Romance locales do not define, and the derived guard already
exempts them — but only when the family's `.Other` form exists in English, so the exemption cannot
swallow a typo in the stem.

## Verification

- 8 061 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 34 Scheduler E2E tests green against a fresh Release docs build.
- Disable-check: suppressing the popover and sharing the popover id each redden their own test.
- No line-ending churn against master.

@codex review

Brain2k-0005:
## Round 5 — all four applied, and to the original as well

All four are in `SchedulerMonthView` too — I ported its popover faithfully, defects included — so
they are fixed in both. Leaving a knowingly broken twin beside a fixed copy is worse than the wider
diff.

- **The registration names its trigger.** Without it the document-level `mousedown` reads a second
  press as an outside click and closes before `@onclick` runs, so the click reopened what it meant
  to close and the button never closed anything.
- **Dismissal unregisters itself.** Clearing only the flag left the JS handler map and the interop
  dictionary holding a callback for a dialog that had left the DOM — and with the flag cleared, no
  later close path could reach it.
- **Escape is handled on the trigger.** A keyboard user who opens with Enter still holds focus
  there, where the dialog's own keydown hears nothing. The trigger also carries `aria-expanded`
  now, since it turned out to be announcing nothing about the thing it controls.
- **The last two columns open inward.** A 14rem panel off the final day column lands past the
  grid's right edge inside a scroller that clips.

## What the alignment test caught on the way

My first version wrote the ternary inline in the `class` attribute. It compiled, rendered, and
produced an **empty class** — Razor does not survive the nested quotes there. The test failed on an
empty string rather than on the wrong alignment, which is the only reason I noticed.

Both views now pick between whole literal strings in a helper, which is what the Tailwind v4 build
needs regardless: it only emits utilities it can see in source.

That is the second Razor-markup subtlety this PR has turned up — the first was the comment inside
an opening tag, which the repo's own lint guard caught.

## Verification

- 8 065 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 34 Scheduler E2E tests green against a fresh Release docs build.
- Disable-check per fix, one at a time; all four go red with only their own fix reverted.
- The month view's existing popover tests stayed green through its four.
- No line-ending churn against master.

@codex review

Brain2k-0005:
## Round 6 — all five applied, plus the one scrollbar the panes were missing

**The MultiDay crash is real and now covered.** A ten-day window with an all-day event on its ninth
day threw `'8' cannot be greater than 7` before it rendered anything — the all-day strip packs
through the month row packer, hard-wired to seven columns. The packer takes the row's real width
now, and its clamps can no longer invert. The test pins both halves: it renders, *and* two events
on days nine and ten share a lane, which is what packing against the true width buys — the
seven-wide packer folded every day-eight-or-later start onto one column and read them as
overlapping.

**The overflow dialog scrolls.** Fair: the cap that hides events created a list that could run past
the bottom of a clipping scroller while the sticky dialog stayed pinned.

**Teardown releases an open registration.** Switching views tears the grid down, dialog and all.

**The month view aligns by visible ordinal.** With `HideWeekends` on a Monday-first grid, Friday is
the last column on screen at raw index 4 — a raw test left it opening outward past a five-column
grid that clips.

**Both views use logical insets.** `start-0`/`end-0` rather than `left-0`/`right-0`, so the index
test stays correct when RTL places day 0 at the right.

## Also in this round: one scrollbar for every pane

Reported against the running docs — each pane had its own scrollbar, so scrolling one calendar left
the others behind. Two scrollers were doing it: the wrapper wrapped every pane in an `overflow-auto`
box, and each view is a scroll container in its own right. The per-pane box is gone, and the three
views that scroll take a `SelfScrolling` parameter the wrapper turns off while panes are up.

The sticky day header survives it because its offset is relative to the nearest scrollport, which is
now that shared box — so every pane's header lines up along the top edge, which is better than what
it replaced.

**Worth recording:** my first pass fixed only the views, and the bUnit test went green because its
predicate looked for `overflow-y-auto` while the wrapper's box used `overflow-auto`. The browser
found the two scrollbars the test could not see. The predicate covers both spellings now.

## Verification

- 8 074 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 34 Scheduler E2E tests green against a fresh Release docs build.
- Browser check on the panes: exactly one scrollable box, and one wheel gesture moves both panes by
  the same amount.
- Disable-check per fix. Two of my first probes were wrong rather than the tests — one removed a
  `CloseAllDayMoreAsync()` call from the wrong method, the other reverted an argument when the
  crash was actually fixed by the clamp. Re-probed at the right sites, all discriminate.
- No line-ending churn against master.

@codex review

Brain2k-0005:
## Round 7 — all five applied

**The sticky header genuinely did not stick in pane mode.** You are right, and my previous commit
message claimed otherwise — I had verified stickiness in the overlay demo and assumed it carried.
The root's `overflow-hidden` clips to the rounded border *and* makes each root a scroll container,
which is what sticky is positioned against; with the scrolling moved out, that container never
moves. `overflow: clip` clips identically without being a scroll container.

Verified in a browser this time: the content moves 542px, the header 42 and then stops, level
across both panes.

**The resource view had the same problem from the other side.** Its headings stayed put precisely
*because* they sat outside the body scroller — sharing the scroll made them ordinary content. They
are sticky now, and its horizontal scroller no longer claims the vertical axis with it.

**Panes reserve a common all-day floor.** This is the one that would have quietly undone the
feature: the panes move by the same pixels, so a differing offset never closes and the same hour
sits at two heights. The floor is an upper bound — it counts all-day *events* where the strip packs
overlapping ones into shared lanes and knows the visible range — because equal offsets matter more
here than a tight fit, and the view clamps it to its own maximum anyway.

**A refresh that removes the overflow closes its dialog**, rather than leaving the DOM with a live
registration nothing can invoke.

**The Arabic dual is gone.** `LumeoLocalizer`'s plural rules are documented as CLDR-lite and return
only `One`/`Few`/`Many`/`Other`, so `.Two` was unreachable the moment I wrote it. Teaching the
pluraliser the dual would change every Arabic count in the library, which is not a decision this
PR should be making.

## One thing I left

The per-pane label row — the calendar's name and its merge button — still scrolls away with the
content. The calendar chips above the scroller stay put and the panes keep the chips' left-to-right
order, so the columns remain identifiable. Making it sticky too would need its height to offset the
day header's, which nothing here can measure. Say the word if you would rather have it.

## Verification

- 8 078 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 34 Scheduler E2E tests green against a fresh Release docs build.
- Browser: one scrollable box in the pane set, both panes move together, both headers stick level.
- Disable-check per fix; all four code changes go red with only their own reverted.
- No line-ending churn against master.

@codex review

Brain2k-0005:
## Round 8 — the scope was wrong, and that was most of the round

Four of the six findings — the resource headings, the resource all-day band, the shared resource
band heights, the timeline row heights — exist only because I extended the shared scroll to **every**
view. They are all the same observation from different angles, and it is right.

A shared vertical scroll means something when the axis is the **same** in every pane. The clock is.
A Resource pane's rows are its own resources; a Timeline's row heights come from its own overlaps;
two agendas of different lengths share no row N. Moving those panes together aligns nothing — and
costs each of them the headings its own scroller was keeping in place.

So time grids share, and everything else keeps its scroller. `SchedulerResourceView` and
`SchedulerAgendaView` are back to exactly what they were, sticky attempt and all: the sticky
heading could never have worked there anyway, because `overflow-x-auto` with `overflow-y-clip`
computes the clip to `hidden` — thank you for that one, I had the spec rule backwards.

**The P1 was mine and it was a crash.** The floor is a switch now rather than a count: as a number
it could exceed the view's own cap and reach `Math.Clamp(value, 4, 3)`, whose minimum above its
maximum throws. It also spares the wrapper a count it cannot make correctly — it sees source events
where the strip packs overlapping ones into shared lanes and expands a recurrence into several,
which is your second finding and disappears with the same change.

## On the failing check

The E2E job on the previous head was **cancelled during "Install Playwright browsers"**, not failed
on a test. The suite is green locally against a fresh Release docs build, and this push re-runs it.

## Verification

- 8 080 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 34 Scheduler E2E tests green locally against a fresh Release docs build.
- Browser: one scrollable box in a Week pane set, both panes move together, both headers stick
  level.
- Disable-check: a floor above the cap reddens the crash test; sharing the scroll for every view
  again reddens the scoping test. My first version of that second test did **not** discriminate —
  it counted scrollers anywhere, and these views scroll themselves regardless, so the pane box is
  what it had to assert on.
- No line-ending churn against master.

@codex review

Brain2k-0005:
## Round 9 — all seven applied

**The pane names are out of the scroller.** I had called this an open trade-off in round 7 and
argued the chips carry the identity; you are right that they do not — the sticky day rows are
*identical* in both panes, so once the names scroll away nothing distinguishes the columns. The
name row now sits above the scroller and keeps its column geometry.

**The reserved strip holds the overflow row too.** Good catch: normalising `[data-lane]` counts
left the "+N more" trigger as an extra row in the overflowing pane only, so the floor still failed
at the one thing it exists for.

**The popover is keyed by date as well as index.** Navigation rebuilds the columns underneath it,
so an index that still overflows in the new range kept the popover open over another day's events.
Its state is cleared synchronously in the rebuild and the registration is reconciled on the next
render — rather than firing the async close and dropping its task, which is the same comment from
the other direction.

**Escape returns focus to the trigger**, in both views, before the close removes the focused
element.

**`ReserveFullAllDayStrip`'s public summary says what it does.** Fair — the XML doc flows into the
registry and the MCP description, and implementation history is not caller-facing. The reasoning
moved to a code comment; the generated description no longer mentions a review or a rejected API.

## One the review did not raise

Removing `SelfScrolling` from the resource and agenda views left the wrapper still passing it. That
does **not** fail: `CaptureUnmatchedValues` swallowed it and rendered `SelfScrolling="False"` as a
stray HTML attribute. Found while re-reading the call sites for this round.

## Verification

- 8 084 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 34 Scheduler E2E tests green against a fresh Release docs build — and the e2e check passed on the
  previous head too, so the earlier red was the cancelled Playwright install.
- Browser: one scrollable box, the day header now holds at 283px while the content moves 501, level
  across both panes.
- Disable-check per fix; all discriminate. The date-key probe needed a second attempt — my first
  `sed` failed on the pattern and reported a pass that never ran the revert.
- Generated MCP description checked for the review reference: zero hits.
- No line-ending churn against master.

@codex review

Brain2k-0005:
## Round 10 — all three applied

**The gutter is reserved on both sides now.** This is the third time in this PR that something laid
out *beside* the scroller lost the scrollbar's width and stopped lining up with what was under it:
the day header in round 1, the all-day strip in round 3, and now the name row I moved out in round
9. The first two I fixed by moving the element inside. This one cannot move — that is the whole
point of it — so both sides reserve the gutter explicitly instead. Measured in a browser: both name
columns sit exactly over their panes, 0px off in x and width.

**The inward threshold scales with the column count.** Right: the panel is a fixed 14rem against
columns of `gridWidth / Days`, so it covers more of them the more days are shown. Derived against a
~900px grid, which gives the same two columns at seven days and four at fourteen — the theory has
all four cases, both sides of the boundary at each width.

**The early return does the reconciliation too.** A refresh that removes *every* all-day event goes
through that path, and it left the dialog's state and registration behind.

## Where this PR stands

Ten rounds, 43 findings. Worth saying plainly: since the shared pane scroll went in at round 6,
most findings have been consequences of my own previous fix rather than of the original defects.
The three problems actually reported — the missing day header, the clipped first hour, the untranslated
dialog — have been done and verified since round 1, and the one scrollbar per pane set works.

If you would rather land the reported fixes now and let the pane-scroll work continue separately, I
am happy to split it; say the word. Otherwise I will keep going.

## Verification

- 8 090 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 34 Scheduler E2E tests green against a fresh Release docs build.
- Browser: name row and panes aligned to the pixel; one scrollable box; headers stick level.
- Disable-check per fix; all three discriminate.
- No line-ending churn against master.

@codex review

Brain2k-0005:
## Round 11 — applied, by removing the estimate rather than retuning it

You are right, and my round-10 fix only moved the constant. Counting columns cannot work here at
all: the panel is a fixed 14rem while a column is `gridWidth / Days`, which changes with the day
count **and** with the width — and side-by-side panes halve the width without touching the count,
so a formula tuned for one is wrong for the other.

The popover is anchored to the **strip** now, not to a day cell. There is nothing left to estimate:
the near half opens from the start edge, the far half from the end, and `max-w-full` handles a grid
narrower than the panel. `SchedulerMonthView` carried the same arithmetic and gets the same
treatment against its grid.

## The test took two goes to become honest

My first E2E version asserted only that the panel sits inside its pane — which the **broken**
version satisfies. Anchored to a cell, `max-w-full` resolves against a 40px day column, so the
panel collapses to fit instead of overhanging. It passed against the bug.

It now also asserts the panel keeps its own width, and runs at a 900px viewport, because a wide one
leaves a half-width pane roomy enough for either anchoring. Reverting the anchor fails it with
`the panel is only 40.0625px wide - it is being measured against a column`.

Seeding the harness for it also broke a neighbouring test: six all-day events landed on the same
day as the two timed pane events, so Month view hid them behind the overflow trigger. They have
their own day now.

## A flaky test you may want to know about

`TooltipClickDismissAndArrowAnchorTests.After_A_Mouse_Click_The_Tooltip_Stays_Closed_Until_The_Pointer_ReEnters`
fails roughly two runs in three on **net8.0** and passes every time on net10.0. I reproduced it on
master at `34b77d56`, so it is not from this branch — but a full-suite run reports it beside
whatever you are working on, which is how it looks branch-specific. Happy to open an issue if you
want one.

## Verification

- 8 090 tests green on net10.0; on net8.0 the same 8 090 minus that pre-existing tooltip flake.
- 35 Scheduler E2E tests green against a fresh Release docs build.
- Disable-check: reverting the anchor fails the E2E test on width; reverting either half-rule fails
  its own theory case.
- No line-ending churn against master.

@codex review

Brain2k-0005:
## Round 12 — applied, and it was worse than the finding said

You are right that `CellClasses` keeps the cell positioned, so the grid never became the containing
block. What that means in practice is that my round-11 change made the month popover **48px wide**:
`max-w-full` resolved against one day column. The class-only test passed the whole time.

Anchoring to the grid would not have worked anyway, which I should have checked before copying the
approach across. The time grid's strip is **one row**, so a panel measured against it still lands
under its trigger. The month grid is six rows deep, and `top-full` against it drops the panel below
the entire month.

So the month keeps the cell as its containing block and loses `max-w-full`. The half rule alone
keeps the panel inside the grid down to about 392px — twice its own 14rem less a column — which a
half-width pane clears at any sane scheduler width.

The month popover now has the browser test the time grid's got, and for the same reason: width,
inside its pane, and opening directly under its trigger rather than below the grid. Putting
`max-w-full` back fails it with `the panel is only 48.0625px wide`.

## On the red check

The `e2e` run on the previous head was **cancelled**, not failed — same as two heads ago. Locally
all 36 Scheduler E2E tests pass against a fresh Release docs build.

## Verification

- 8 090 tests green on **both** TFMs this time — the net8.0 tooltip flake did not reappear in this
  run, which is consistent with it being intermittent rather than fixed.
- 36 Scheduler E2E tests green.
- Disable-check: restoring `max-w-full` fails the new month test on width.
- No line-ending churn against master.

@codex review

PR #428, opened 2026-08-19
feat(scheduler): one clock for a pane set, and a tooltip that finds its event

Brain2k-0005:
Two things reported against the running docs.

## A timed event's tooltip opened at the bottom of the page

The pill is placed **absolutely** by the clock, so the wrapper `Tooltip` renders around it
collapses to a zero-size box at its *flow* position — measured on the live site, `0x0 at y=1440`
for a pill drawn at `y=702`. floating-ui was positioning the box against that, faithfully.

`Tooltip` takes an optional `AnchorElementId` now, defaulting to its own wrapper, and the time grid
points it at the pill. The month and all-day pills never had the problem — they sit in normal flow,
so the wrapper already has their geometry, which is why the month view's existing tooltip test
stayed green throughout.

`TooltipContext` carries the anchor as a **body property**, not a positional one: the record is
shipped, and adding to its primary constructor would change the synthesized signature and
`Deconstruct` for every existing caller.

## Side-by-side panes repeated the same hours in every column

They read against one clock now, on the leftmost pane, the way Outlook shows them.

The pane carrying the axis is wider by exactly the gutter, and its flex basis gives that width back
so every pane's **day columns still come out the same size** — measured at 390px each. Without
that, the calendar with the clock would be the narrowest one, which defeats standing them side by
side. The name row above matches the same geometry, with its first label padded past the hour
labels so it reads above the days rather than above the clock.

## Verification

- 8 096 tests green on both TFMs (7 new), `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 38 Scheduler E2E tests green, 2 new — both of these are geometry, which markup assertions cannot
  see.
- Browser, against a fresh Release docs build: one hour gutter for the pane set, both day grids
  390px, name row aligned to the pixel, and the tooltip 36px above its pill where it used to be
  241px below.
- Disable-check per fix; all four go red with only their own reverted.

@codex review

## Summary by CodeRabbit

- **New Features**
  - Added support for positioning tooltips against a specific anchor element.
  - Added control over time-gutter visibility in scheduler views.

- **Improvements**
  - Improved tooltip placement for timed scheduler events, including dynamic anchor changes and retry handling.
  - Side-by-side scheduler panes now share a properly aligned time gutter and scrollbar space.
  - Preserved grid alignment when the time gutter is hidden.

- **Tests**
  - Added coverage for scheduler pane alignment, gutter behavior, unique event identifiers, and tooltip positioning.


Brain2k-0005:
## Round 1 — applied

Both of you found the same thing, and it is a fair one: I added `AnchorElementId` without adding it
to the guard that decides when `PositionFixed` re-runs — the guard that already tracks `Side`,
`Offset` and `Align`, for exactly the same reason. A caller moving the anchor while the tooltip was
open saw nothing happen until it closed and reopened.

`_appliedAnchor` sits beside the others now, is included in the re-placement condition, and is
cleared with them on cleanup.

Two tests: moving the anchor re-places against the new element, and an *unchanged* anchor does not
re-place on every render — which is what the applied-value bookkeeping exists for in the first
place, and would have been easy to break while fixing the first half.

## Verification

- 8 098 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 40 Scheduler and Tooltip E2E tests green against a fresh Release docs build.
- Disable-check: dropping the anchor from the guard reddens the move test.
- No line-ending churn against master.

@codex review

Brain2k-0005:
## Round 2 — applied, and the fix needed a second layer to work at all

You are right, and the test I wrote for it found that my first attempt was inert.

`positionFixed` now returns an **empty string** when the reference element is missing, and
`TooltipContent` leaves its applied state alone when it sees one, so the next render tries again.
Guarded rather than returned early — the exit wiring at the end of that method has to run on every
pass.

**That alone changed nothing**, which the test caught. `FloatingPositionInterop` was replacing any
`null`-or-empty result with the requested side, on purpose: a stale cached script returns nothing,
and a directional-arrow consumer still needs a sensible value. It swallowed the new signal too. The
two cases are distinguishable — `undefined` deserialises to `null`, `''` stays `""` — so it now
falls back only on `null`. Only `Tooltip` reads this return value; every other caller awaits and
discards it.

## And last round's no-thrash test was vacuous

Re-rendering the host with an unchanged anchor does not re-render the content at all, so "no second
placement" was free rather than earned. It only surfaced because *this* round's test failed for the
same reason and I instrumented it — `afterOpen=1 now=1`, no second render at all.

Both tests now force a real re-render through a parameter the placement guard does not read.

## On the red checks

- `build-and-test` failed on `GanttV3KeyboardNavigationTests.Mid_Flight_Combination_Drift`, the
  known flake tracked in #413. Not in anything this branch touches.
- `e2e` was **cancelled** again, not failed.

Both re-run on this push.

## Verification

- 8 099 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 40 Scheduler and Tooltip E2E tests green against a fresh Release docs build.
- Disable-check: latching regardless of the result, and restoring the interop's null-or-empty
  collapse, each redden the retry test on their own.
- No line-ending churn against master.

@codex review

Brain2k-0005:
## Round 3 — all three applied

Three consequences of the retry fix, all in the same seam, and all fair.

**The script gave up before cleaning up.** A box whose anchor was swapped for an element not yet in
the DOM kept being positioned against the element it used to follow, because the early return sat
above the teardown. It tears the previous placement down first now, then reports that it placed
nothing — which also makes the empty return honest: there is no stale watcher hiding behind it.

**The side-change handler outlived a placement that never happened.** It is stored *before* the JS
call, deliberately, because a flip can be reported during it. When nothing is placed there is no
watcher to report anything, and a surface torn down before a retry succeeded left the entry — and
the component it closes over — for the lifetime of the service. Removed again on an empty result.

**And the contract now says what it does.** You are right that returning `""` while the interface
documented "always the resolved side" leaves the doc lying to every implementer. It documents the
empty case explicitly: what it means, that any previous placement is gone, and what each kind of
caller should do — record nothing and retry, or leave a directional arrow where it is. I preferred
that to a second method, because there is exactly one consumer of this return value and a parallel
API would have to be kept in step with it forever.

## Verification

- 8 100 tests green on both TFMs, `0 Error(s) / 0 Warning(s)` on a clean rebuild.
- 40 Scheduler and Tooltip E2E tests green against a fresh Release docs build.
- Disable-check: removing the handler cleanup reddens the new test, which drives a real flip through
  `OnPositionSideChanged` and asserts it reaches nobody.
- The previous round's red checks were the #413 flake and a cancelled e2e; both passed on this head
  before these changes.

@codex review

PR #429, opened 2026-08-19
fix(scheduler): one scrollbar for month panes, with their weeks in step

Brain2k-0005:
Reported against the running docs: three calendars side by side in month view, three scrollbars.

## What was wrong

The shared-scroll rule added for the time grids in #427 listed Week, Day and MultiDay only. The month grid was left out by omission, although its vertical axis is the same dates in every pane - one scroll offset means the same week everywhere.

Sharing the scroll turned out to be half the fix. A week is only *at* the same height in every pane if it *is* the same height, and a pane whose day overflowed carried a `+N more` row the others did not, so that week stood 26px taller and every week below it sat out of step for the rest of the month.

## What changed

- `SchedulerMonthView` takes `ReserveFullLanes`, which holds every week open at its full lane budget including the row the trigger occupies. The wrapper turns it on for panes; a single calendar still sizes each week to its own content.
- `SharesTimeAxis` splits off from `SharesPaneScroll`. The hour gutter is a time-grid concept - the month grid has no clock, and handing its first pane the gutter's width would only make that calendar wider than the others.
- A filled lane and an empty one now come from one named constant. Left to size itself a pill was 17.75px against an empty lane's 19.
- The reserved overflow row mirrors the trigger's markup instead of approximating its height. A bare span was 16.5px against the real row's 24 - the wrapper's line box, not the button, sets that.

The last two were found by measuring the rendered page, not by reading the markup: after the first fix the panes still drifted 3.75px per row.

## Verification

- Clean rebuild: 0 errors, 0 warnings.
- 8,103 unit tests green on net10.0 and net8.0.
- E2E suite run the way CI does (docs on 5290 + test server host on 5299): 275/277. The two failures - the home-page visual baseline and a Gantt wheel-zoom test - reproduce identically on this branch with the source changes stashed (same 86,292 pixels, same 752px drift), so they are pre-existing and local to this machine.
- Measured on the running docs: 1 scrollable box, and week rows at 314/436/558/680/802/924 in *both* panes.

Three markup tests and one E2E test cover it; each was disable-checked, and the E2E one was re-checked after the first attempt turned out to be measuring a stale server.

@codex review

## Summary by CodeRabbit

* **New Features**
  * Added an optional setting to reserve consistent event lanes across all weeks in Month view.
  * Improved side-by-side Month views with shared scrolling and aligned week rows.
  * Prevented unnecessary time-axis spacing in Month view panes.

* **Bug Fixes**
  * Month panes now maintain consistent heights when only some panes contain overflowing events.
  * Improved overflow-row sizing and safely handled negative lane limits.

* **Tests**
  * Added coverage for shared scrolling, pane alignment, lane sizing, overflow behavior, and edge cases.


Brain2k-0005:
Fixed all five findings from the first round.

- **Reserved overflow row could wrap** (Codex, P2) — the real trigger had no fixed height and no `nowrap`, so a longer localized label would have made that week two rows tall in one pane and one in the next. Both the real and the reserved row now come from one pair of constants: a fixed-height wrapper and a single-line truncating label. Measured at 1400/1000/820px: exactly 24px in both cases, rows still aligned.
- **Fixed pill height broke custom styling** (Codex, P2) — right, and it hit plain single calendars too. The lane height is a *floor* now (`min-h`), so `ResolveEventClassNames` can still make a pill taller instead of spilling into the lane below. The alignment case is unaffected: a default pill is shorter than the floor either way.
- **Vacuous lane assertion** (Codex, P3) — correct, and the more useful half of it: `[data-lane]` was never emitted by the month view, so the counts were `0 == 0` and the test passed with the fix reverted. Month lanes now carry the marker its sibling time-grid view already used, and the assertion compares every cell rather than a sum, guarded against the all-zero case.
- **Negative `MaxVisibleLanes` reached an array length** (CodeRabbit, Major) — clamped, with a test.
- **`ReserveFullLanes` description mixed in shared scrolling** (CodeRabbit, Minor) — the summary now leads with what the parameter does; the pane rationale follows as one sentence. Registry regenerated.

Three tests added and one repaired. Each was disable-checked — including a second pass on the repaired one, because my first mutation kept the lane count *uniform* and so did not discriminate.

Gate: clean rebuild 0/0, 8,106 unit tests green on net10.0 and net8.0, E2E 275/277 with the two known pre-existing failures.

@codex review

Brain2k-0005:
Round two, both findings fixed.

**Negative budget reached `HiddenCounts`** — correct, and the more interesting half: clamping at the allocation only meant the hidden count still subtracted the raw value, so an empty calendar showed a `+3 more` in every cell with nothing behind it. One clamped budget is computed up front now and both readers take it. The existing test was extended to assert no trigger appears.

**Custom pill heights vs. cross-pane alignment** — this one and the P2 from round one ask for opposite things: a fixed height clips a pill that `ResolveEventClassNames` made taller, a floor lets it open its lane in one calendar and not the next, which is the drift reserving exists to remove. Deriving the reserved height from the rendered lanes across all panes would mean measuring every pane at runtime and feeding a height back into the others — a lot of machinery, and a new class of layout race, for a case this narrow.

So it is decided per mode: a floor normally, a hard height while reserving. Outside pane mode nothing is constrained and a custom pill grows exactly as before, which was the round-one objection. Inside it, alignment wins and the lane clips — in a mode the caller opted into for that geometry, and only for pills styled beyond the lane height. If you would rather have the runtime-measured version, say so and I will build it, but I think the trade is the right way round here.

Both directions are pinned: `A_filled_lane_takes_the_lane_height_as_a_floor_not_as_a_ceiling` and `While_reserving_the_lane_is_a_hard_height_because_alignment_outranks_a_taller_pill` each go red with only their own half reverted.

Gate: clean rebuild 0/0, 8,107 unit tests green on net10.0 and net8.0. Local E2E showed a third failure this round in `DragInteractionTests.Dragging_the_splitter_separator_moves_it` — it passes 3/3 in isolation and touches Splitter, not Scheduler, so it reads as load-sensitive here; CI's own e2e job is the check that counts.

@codex review

Issue #436, opened 2026-08-31
Field report: Lumeo 5.0.0 in a production Next.js→Blazor migration (40 findings)


Brain2k-0005:
An external team (DocFlow) rebuilt a production Next.js 15 UI — shadcn/ui + TanStack Table + an embedded ReUI DataGrid — 1:1 in Blazor WebAssembly with Lumeo 5.0.0, aiming for pixel-level parity. The app is a dense receipts grid (250–1000 rows), a detail dialog and 16 further list/detail routes. Every entry was checked against a running app, not read out of the source.

This issue tracks the report as a whole. The findings I judged most consequential are filed separately:

- #431 — `OnRangeRequest` never fires (§1.1a) — **root cause found**, branch order in `DataGridBody.razor`
- #432 — `PageSize` eagerly materialises skeleton rows, seconds of main-thread freeze (§1.1c) — **root cause found**
- #433 — `Lumeo.DataGrid` always pulls the Excel/PDF stack, 1.44 MB brotli (§6.1)
- #434 — no density/size tokens as CSS variables (§2.1)
- #435 — `lumeo-utilities.css` covers only Lumeo's own utilities (§2.4)

## What they said works

Worth recording, because it is the part that made the migration viable at all: column resize with auto-fit, drag reorder, pinning, column chooser, layout persistence including sort and filters, server-side sorting, footer aggregates, fullscreen and the built-in German translations all worked first try. They call the colour-token compatibility with shadcn "herausragend" — their `globals.css` variables carried over with no adaptation, and they name that as the main reason 1:1 parity was reachable. They single out 4.1 → 5.0 as a clear step forward, particularly `CultureInfo.CurrentUICulture` now actually driving the built-in strings.

## Remaining findings, not yet filed

[...]

PR #459, opened 2026-09-03
5.8.0: Filters, a filter builder for Lumeo.DataGrid

Brain2k-0005:
## What

`Filters`, the filter builder the DocFlow field report asks for in §15 (#436): ReUI's Filters ported to `Lumeo.DataGrid`, basic variant with async options. Stages 1 and 2 of the report's plan; the advanced variant (groups, and/or, drag) follows in its own PR. Version 5.8.0 on master, unreleased.

### The bar

- **Add a filter** opens a searchable attribute list (a `Command`); nested attributes are branches the list drills into, with a breadcrumb and a search that reaches every level. Picking an attribute adds a chip with no condition and opens its condition menu; picking the condition opens the value editor; Enter or a pick commits. An unfinished rule stays as a chip and does not filter (`Flatten()` skips it).
- **A chip** is field | condition | value | menu, fused by hand (each trigger sits in its own popover wrapper, so a ButtonGroup cannot). The condition and value segments reopen on click; the menu duplicates, negates (swapping to the declared inverse, `is` to `is not`, else flagging the rule) or removes. A rule whose field left the schema renders as a removable chip.
- **Seven value kinds** with ReUI's operator catalogue: text, number, range, date (with `is before`, `is after`, `is on or before`, `is on or after`), select, multiselect (`has any of`, `has all of`, `has none of`), boolean. An editor per kind: text and number inputs with fused apply/discard buttons, a number range with a footer, a date editor that parses "today", "next week", "3 days ago", a weekday or a culture date and shows a calendar (a range shows the range calendar), option lists with search, check marks, pinned selected options and an exclusive "none" option, a two-row boolean list.
- **Async options**: `LoadOptions(query, cursor, token)` with a 250ms debounce, cancellation of the superseded request, paging through a load-more row, an error row with retry. `ResolveValues` gives a chip restored from a saved query its labels before the list was ever opened; a shared resolution cache means every chip and editor of the bar sees what any of them loaded.
- **Customisation**: `Editors` registry by key, a field's `EditorTemplate`, `ValueTemplate`/`ValueText` per field or for the whole bar, `ChipTemplate`, `Trigger`, `Labels` (every string, from the `Filters.*` locale keys, English and German plus the twelve other shipped locales), `MenuClass`, `MaxPathSegments`.
- **States and access**: `Size` Sm (28px) or Md (32px) on the geometry tokens, `Disabled` (no tab stops), `ReadOnly` (focusable, `aria-disabled`, refuses edits), `ShowClear`; a live region announces counts, groups and exclusive clears; roving tab stop across chips, arrows/Home/End move, Delete removes, Enter opens, Alt+arrows reorder.
- **Model**: `FilterQuery` (root group) of `FilterRule` (Path, Operator, Value, Negated) and `FilterGroup` (Combinator, Rules), immutable records; `FilterQueries` holds every operation as a pure function that returns the same instance when nothing changed; `Flatten()` returns `FilterTerm`s for an API; `CollectIssues` reports what is not ready. `FilterField`, `FilterChoice`, `FilterOperatorDef`, `FilterLabels`, `FilterDateValue`, `FilterOptionsState`, `FilterEditorContext`, `FiltersContext` for custom parts.
- `Command` gains `SearchChanged` and `ShouldFilter` (cmdk's `shouldFilter`) so a list can filter or load its items itself.

### Docs

`/components/filters` with default, small, nested, async, custom editor and value, read only and disabled, events and veto demos; the props table from the registry; the operator catalogue per kind.

## Verification

- Clean Release build of `Lumeo.slnx`, Lumeo.Tests (both TFMs), Lumeo.Docs.Tests, Lumeo.Cli.Tests, docs parity, RegistryGen.
- `FilterModelTests`: the query operations (insert, remove, duplicate, update, move, wrap, unwrap, prune, issues, flatten), the catalogue, default and negate, coercion, the index with nested paths and deep search, the exclusive rule, date parsing and formatting, labels.
- `FiltersTests`: preset chips, the add flow through picker, condition and value, select commit, menu negate/duplicate/remove, clear with its announcement, veto, read-only and disabled, Delete on a focused chip, a controlled query, an unknown field, no value segment for a no-value operator.
- Headless on the docs page: chips measure 28px (Sm) and 32px, ends rounded and inner segments square; the add flow ends with focus inside the condition menu, then inside the value input; Enter commits; a multiselect commits without closing; negate swaps the operator; a restored async chip reads "Person 42" after `ResolveValues`; the async list pages 20 rows plus load more. No console errors.


Brain2k-0005:
The review findings on this PR are fixed in #460 (the advanced variant itself, the chip's field segment, DefaultOperator and Validate wiring, Reorderable for Alt+Arrow, JSON round trip, the UpdateRule no-op, localized date phrases, async request invalidation, restored async selections, inline list max-heights, CommandGroup under ShouldFilter=false, labels after a culture switch).

PR #493, opened 2026-09-16
Gantt (legacy): fix progress-fill contrast regression in dark mode

Brain2k-0005:
## Summary

Owner report on the live site after PR #490 (label legibility fix) merged: the legacy `Gantt` (v2) progress fill regressed into the INVERSE of the original defect in dark mode - a near-white opaque-looking block with white label text on it, invisible.

**Root cause (verified live, exact pixel samples below):** PR #490 changed the v2 progress rect from fully opaque to `fill-opacity: 0.55`. That was still too dark, because Lumeo's default "zinc" scheme sets `--color-primary` to the *same value* as `--color-foreground` in each mode - a monochrome "ink" primary (near-black in light mode, near-white in dark mode). The Gantt's default bar colour (no `BarColor` set - the common case) is `--color-primary`, and the label text is `--color-foreground`, so at 0.55 opacity the fill converged close enough to the label's own colour to lose contrast. Confirmed live: in dark mode both tokens read exactly `oklch(0.9848 0 0)`.

I could not reproduce the second reported symptom (light mode "washed out", strong dark zebra bands, black header) with the customizer's mono font preset alone, the full "Lyra" preset, or any combination - screenshots attached below all read clearly. `git diff` of PR #490's merge commit against its parent touches only `Gantt.razor`, `GanttBar.razor`, `GanttChart.razor` and `gantt-v2.js` - no row-band/header/zebra colours or SVG-root/group opacity changed. I don't believe PR #490 caused that symptom; flagging it back to the owner for a fresh repro (exact preset/scheme, browser, zoom) since I can't chase it further without reproducing it.

## Fix

Dropped the v2 progress rect's `fill-opacity` from 0.55 to 0.35 - matching the percentage `GanttBar.razor`'s v3 renderer already uses for the identical translucent-tint mechanism (`color-mix(in oklab, {colour} 35%, transparent)`). `fill` itself is untouched (still the resolved bar/primary colour, never the foreground - custom `BarColor` still applies). Nothing else in the SVG (bgRect, arrows, milestone diamond, grid) changed opacity.

## Verified (live, pixel-sampled)

Enterprise Schedule, "Berth A - MSC Aurelia" (80% progress), 1440x900:

| Mode | Progress-fill RGB | Label RGB | Contrast |
|---|---|---|---|
| Dark, before (0.55) | `rgb(174,174,174)` | `rgb(250,250,250)` white | **~2.1:1** (label nearly invisible) |
| Dark, after (0.35) | `rgb(140,140,141)` | `rgb(250,250,250)` white | **3.2:1** (clearly legible) |
| Light, after (0.35) | `rgb(128,128,130)` | `rgb(9,9,11)` near-black | **5.1:1** (passes WCAG AA) |

Dark mode's 3.2:1 doesn't hit the textbook 4.5:1 AA bar - that's an inherent ceiling of this theme (primary literally equals foreground there), and pushing opacity lower to chase 4.5:1 would drop the progress fill *below* the untouched remainder's own 0.30 opacity, inverting which segment reads as "more filled." I judged holding progress visibly more prominent than the remainder (0.35 > 0.30) more important than closing the last delta to AA once the fix already turns a ~2:1 invisible label into a clearly-readable ~3.2:1 one - flagging this trade-off for review rather than deciding it silently.

Also checked: `/components/gantt-chart` (v3, unaffected by this change - its default bar colour is `--color-chart-1`, not `--color-primary`, so it never had this collision) and the enterprise Schedule + mono font preset (JetBrains Mono) in both themes - all read clearly.

## Test plan

- [x] `dotnet build Lumeo.slnx -c Release -warnaserror` -> 0 Warning(s), 0 Error(s)
- [x] `dotnet test tests/Lumeo.Tests/Lumeo.Tests.csproj -c Release --no-build -f net10.0 --filter "FullyQualifiedName~Gantt"` -> 776 passed, 0 failed (unchanged - this renderer's SVG output isn't part of Blazor's render tree, so there's no existing or new bUnit coverage for the JS-internal fill-opacity constant; verified instead via live pixel sampling as shown above)
- [x] Live check: `/demos/enterprise` -> Schedule, light + dark + dark+mono-font, `/components/gantt-chart` light + dark - screenshots and pixel samples in this task's report


Brain2k-0005:
## Update: added a text halo (both renderers)

Round 2's opacity fix (0.35) got dark mode to ~3.2:1, which isn't enough for the label that's the whole point of the bar - and since `--color-primary` literally equals `--color-foreground` in this theme's dark mode, opacity tuning alone can't fully separate them. Added the standard halo technique instead, which sidesteps that entirely: contrast is measured between the glyph's own foreground colour and a background-coloured ring around it - the one pairing every theme defines as maximally contrasting - independent of the fill underneath.

- **v2** (`gantt-v2.js`, SVG): label gets `stroke: var(--color-background)` (resolved the same way every other token here is), `stroke-width: 3px`, `stroke-linejoin: round`, `paint-order: stroke fill`. Milestone labels untouched (already render outside the diamond). `fill-opacity: 0.35` from the previous commit is unchanged.
- **v3** (`GanttBar.razor`, HTML): in-bar label gets `text-shadow: 0 0 2px var(--color-background), 0 0 2px var(--color-background)` (two identical shadows, not one - reads more evenly around the glyph than a single blur), applied unconditionally so the default colour gets the same safety net a custom `BarColor` already had.

### Re-sampled (halo ring vs glyph ink, sampled from actual rendered pixels within the glyph strokes - not the bar fill)

v2, "Berth A - MSC Aurelia" (80% progress), both segments:

| Mode | Halo-ring pixel | Glyph-ink pixel | Contrast |
|---|---|---|---|
| Dark, progress segment | `rgb(9,9,11)` | `rgb(250,250,250)` | **19.06:1** |
| Dark, remainder segment | `rgb(9,9,11)` | `rgb(250,250,250)` | **19.06:1** |
| Light, both segments | `rgb(255,255,255)` | `rgb(9,9,11)` | **~19.9:1** |

(Halo-ring pixel values sampled directly between glyph strokes match the page's own `--color-background` pixel exactly, in both segments - confirms the stroke color is the token, unaffected by the fill underneath.)

v3, "Wireframes" (80% progress), dark mode: baseline fill-vs-ink (no-halo-credit) was already 9.9:1/15.5:1 (v3's default `--color-chart-1` never had the primary-equals-foreground collision) - the halo pixel sampled adjacent to a glyph edge measured 12.4-14.0:1, an extra margin on top. Visually confirmed clean/legible in both renderers, both themes, at both 3x zoom and true 1x/100% pixel density - the halo does not read as a blur.

Tests: updated the pre-existing `Default_Primary_Bar_Keeps_The_Existing_Text_Foreground_Class_Unchanged` v3 bUnit test (style is no longer `null` now that every label carries the halo) and added `Every_InBar_Label_Carries_The_Background_Halo_Regardless_Of_Custom_Colour`. `dotnet build -warnaserror` -> 0/0. Gantt filter -> 777 passed, 0 failed.

Screenshots and full pixel-sampling methodology in this task's report.

